PR #16, opened 2024-02-18
feat: Implement MedRecordValue

JabobKrauskopf:


crates/medmodels-core/src/medrecord/values.rs
```diff
@@ -0,0 +1,22 @@
+#[derive(Debug, Clone)]
+pub enum MedRecordValue {
+    String(String),
+    Int(i64),
+    Float(f64),
+    Bool(bool),
+}
+
+macro_rules! implement_from {
```
JabobKrauskopf, line 9: Rust is sooooooo amazing for enabling stuff like this

rustmodels/src/medrecord/conversion.rs
```diff
@@ -0,0 +1,111 @@
...
+        match self.0 {
+            MedRecordValue::String(value) => value.into_py(py),
+            MedRecordValue::Int(value) => value.into_py(py),
+            MedRecordValue::Float(value) => value.into_py(py),
+            MedRecordValue::Bool(value) => value.into_py(py),
+        }
+    }
+}
+
+pub trait DeepFrom<T> {
```
JabobKrauskopf, line 56: Also checkout the magic implemented with `DeepFrom`

FloLimebit (changes requested):
maybe we should add a little description in the readme that the installation of rust is needed if someone wants to install medmodels from source. Quick guide on how to install rust (basically two lines i guess)

JabobKrauskopf:
> maybe we should add a little description in the readme that the installation of rust is needed if someone wants to install medmodels from source. Quick guide on how to install rust (basically two lines i guess)

@FloLimebit I agree but I would suggest that this should be addressed in a different issue/pr (https://github.com/limebit/medmodels/issues/22)

FloLimebit (approved):

PR #33, opened 2024-02-23
test: Write medrecord tests

JabobKrauskopf:


philippkochlimebit (approved):

JabobKrauskopf:
The name of the github action step running the tests is currently `Test with pytest` - This is not accurate, as it is now also testing rust code via `cargo test`. Will fix this in #38

PR #57, opened 2024-04-17
feat: MedRecord serialization, deserialization, saving and loading

JabobKrauskopf:
Used `serde` and `ron` to serialize/deserialize and save/load a MedRecord as a `ron` (rust object notation) file.

Can be achieved using `from_ron` and `to_ron`.


FloLimebit (approved):
Fantastic! Now we've got a file format for our graphs! ❤️

PR #62, opened 2024-04-23
refactor: Adjust return types of medrecord functions

JabobKrauskopf:


medmodels/medrecord/types.py
```diff
@@ -6,3 +6,27 @@
 EdgeIndex = int
 Group = MedRecordAttribute
 Attributes = Dict[MedRecordAttribute, MedRecordValue]
+
+
+def is_medrecord_attribute(value: MedRecordAttribute) -> bool:
```
JabobKrauskopf, line 11: These methods are needed to support `isinstance` calls for python 3.9

JabobKrauskopf:
FYI: I accidentally amended more changes to my first commit than i wanted to and I couldn't be bothered to split them up again (everything is still early in development so who cares) so I instead just squashed all of my changes into one commit 🤷‍♂️

philippgrosser (dismissed):

PR #72, opened 2024-05-02
feat: Add schema validation

JabobKrauskopf:


crates/medmodels-core/src/medrecord/group_mapping.rs
```diff
@@ -113,6 +182,21 @@ impl GroupMapping {
         }
     }
 
+    pub fn remove_edge(&mut self, edge_index: &EdgeIndex) {
```
MarIniOnz, line 185: I do not know if the linting of this is wrong, but it appears as it is all strings, correct me if I am wrong
image

crates/medmodels-core/src/medrecord/group_mapping.rs
```diff
@@ -8,21 +8,26 @@ pub type Group = MedRecordAttribute;
...
             groups_of_node: MrHashMap::new(),
+            groups_of_edge: MrHashMap::new(),
         }
     }
 
     pub fn add_group(
         &mut self,
         group: Group,
         node_indices: Option<Vec<NodeIndex>>,
+        edge_indices: Option<Vec<EdgeIndex>>,
```
MarIniOnz, line 30: Does this assume node_indices or edge_indices are None if not specified? Or do they always need to be indicated?

medmodels/medrecord/medrecord.py
```diff
@@ -787,21 +831,36 @@ def add_group(
 
         Args:
             group (Group): The name of the group to add.
-            node (Optional[Union[NodeIndex, NodeIndexInputList, NodeOperation]]):
+            nodes (Optional[Union[NodeIndex, NodeIndexInputList, NodeOperation]]):
+                Node index or indices to add to the group, optional.
```
MarIniOnz, line 835: The description misses to explain how to use NodeOperation, and refers only to the indices.

medmodels/medrecord/medrecord.py
```diff
@@ -787,21 +831,36 @@ def add_group(
...
+            edges (Optional[Union[EdgeIndex, EdgeIndexInputList, EdgeOperation]]):
                 Node index or indices to add to the group, optional.
```
MarIniOnz, line 837: Same as above.

MarIniOnz (commented):
Incredible PR!
Such an amount of content was difficult to validate for me because of my little knowledge of Rest, but regarding formatting and variable naming, looks correct!

crates/medmodels-core/src/medrecord/group_mapping.rs
JabobKrauskopf: Thats just the GitHub parser doing funny things :)

crates/medmodels-core/src/medrecord/group_mapping.rs
JabobKrauskopf: Nope. There are no default function argument values in rust so every argument has to be specified, in this case by using `None` or `Some(...)`

MarIniOnz (approved):

MarIniOnz (dismissed):

PR #85, opened 2024-05-13
fix: Add pyright, fix typing issues, removed black and formatted code

JabobKrauskopf:


FloLimebit (dismissed):
fantastic changes! type saftety is good safety 👷

PR #93, opened 2024-05-24
feat: adding treatment effect estimation class

MarIniOnz:
Adding treatment effect estimation class, which takes MedRecord, outcomes and treatments and divide the patients in the MedRecord in control and treatment (True and False) categories.


medmodels/matching/matching.py
```diff
@@ -0,0 +1,40 @@
...
+            treated_group (Set[NodeIndex]): A set containing the IDs of patients in the
+                treated group.
+            control_group (Set[NodeIndex]): A set containing the IDs of patients in the
+                control group.
+
+        Returns:
+            None
+        """
+        self.medrecord = medrecord
+        assert isinstance(medrecord, MedRecord), "medrecord must be a MedRecord object"
```
JabobKrauskopf, line 34: Asserting that the medrecord really is a medrecord is unnecessary and already covered by the type system. We do not check for user error only when the user can reach an unwanted state while following our function typing.

medmodels/treatment_effect_estimation/tests/test_treatment_effect.py
```diff
@@ -0,0 +1,369 @@
...
+
+import pandas as pd
+
+from medmodels import MedRecord
+from medmodels.medrecord import node
+from medmodels.treatment_effect_estimation.treatment_effect import TreatmentEffect
+
+
+def create_patients() -> pd.DataFrame:
+    """
```
JabobKrauskopf, line 13: FYI. I would say it is not necessary to include docstrings in functions that are not publicly accessible via the package and where it is already clear what the function is doing

JabobKrauskopf: No need to change this however. Only for future reference

medmodels/treatment_effect_estimation/treatment_effect.py
```diff
@@ -0,0 +1,512 @@
...
+
+        Raises:
+            AssertionError: If treatments or outcomes are not provided as lists, are
+                empty, or if the specified patients group is not found within the
+                MedRecord groups.
+        """
+        assert isinstance(treatments, list), "Treatment must be a list"
+        assert treatments, "Treatment list is empty"
+        assert isinstance(outcomes, list), "Outcome must be a list"
+        assert outcomes, "Outcome list is empty"
```
JabobKrauskopf, line 65: Also unnecessary asserts

medmodels/treatment_effect_estimation/treatment_effect.py
```diff
@@ -0,0 +1,512 @@
...
+        assert patients_group in medrecord.groups, (
+            f"Patient group {patients_group} not found in the data. "
+            f"Available groups: {medrecord.groups}"
+        )
+        assert isinstance(
+            time_attribute, str
+        ), f"Time attribute {time_attribute} not found in the data."
```
JabobKrauskopf, line 72: why not type `time_attribute` as `str` when it is asserted here.

Also the assertion message does not make sense imo

medmodels/treatment_effect_estimation/treatment_effect.py
```diff
@@ -0,0 +1,512 @@
...
+            self.control_false = matched_control & self.control_false
+
+        # The groups have been sorted and the other methods can be called
+        self.groups_sorted = True
+
+    def _is_outcome_after_treatment(
+        self, node_id: NodeIndex, outcome: NodeIndex, max_time_in_years: float
+    ) -> bool:
+        """
+        Determines whether an outcome occurred after treatment within a specified time
```
JabobKrauskopf, line 211: These private functions do not need docstrings as well IMO
(also just for reference, no change required)

medmodels/treatment_effect_estimation/treatment_effect.py
```diff
@@ -0,0 +1,512 @@
...
+        else:
+            nodes_to_add = self.medrecord.select_nodes(
+                node().index().is_in(self.medrecord.group(self.patients_group))
+                & criteria_filter
+            )
+            criteria_set = set(nodes_to_add)
+
+        control_all = criteria_set - self.treatment_all
+        assert (
+            control_all != set()
```
JabobKrauskopf, line 348: Does this work correctly?

Checking for size == 0 would be more verbose IMO

JabobKrauskopf (changes requested):
Amazeballs! Some minor changes requested. 🚀 
And shouldn't the commit be called something like the pull request name?

medmodels/treatment_effect_estimation/treatment_effect.py
MarIniOnz: It does work, as it is checked for in the test. I will change it to len()>0 for better readability.

medmodels/treatment_effect_estimation/treatment_effect.py
JabobKrauskopf: That is completely fair and super nice! Just wanted to say that you are not required to do so but if you want to, go ahead! :)

medmodels/treatment_effect_estimation/treatment_effect.py
```diff
@@ -0,0 +1,507 @@
...
+                patients within the dataset. Defaults to "patients".
+            time_attribute (MedRecordAttribute, optional): The attribute representing
+                time within the dataset. Defaults to "time".
+
+        Raises:
+            AssertionError: If treatments or outcomes are not provided as lists, are
+                empty, or if the specified patients group is not found within the
+                MedRecord groups.
+        """
+        assert treatments, "Treatment list is empty"
```
JabobKrauskopf, line 62: Maybe use `assert len(treatments) > 0` like you did in other cases

medmodels/treatment_effect_estimation/treatment_effect.py
```diff
@@ -0,0 +1,507 @@
...
+        return False
+
+    def find_first_time(self, node_id: NodeIndex) -> pd.Timestamp:
+        """
+        Determines the timestamp of the first exposure to any treatment in the
+        predefined treatment list for a specified patient node. This method is crucial
+        for analyzing the temporal sequence of treatments and outcomes.
+
+        Args:
+            node (NodeIndex): The patient node for which to determine the first treatment
```
JabobKrauskopf, line 260: It should be `node_id`. Also maybe rename all `node_id`s to `node_index` as that is the new and more accurate naming in the MedRecord

JabobKrauskopf (changes requested):
Sry still some very minor adjustments 🙃

JabobKrauskopf (approved):
Amazingsauce with Amazeballs 🍝

JabobKrauskopf:
This branch should no longer be used. Work should continue in #111

PR #96, opened 2024-05-27
Refactor Mahalanobis metric

MarIniOnz:


medmodels/matching/metrics.py
```diff
@@ -63,6 +65,43 @@ def exact_metric(vector1: NDArray[np.float64], vector2: NDArray[np.float64]) ->
     return np.inf
 
 
+def mahalanobis_metric(*vectors: NDArray[np.float64], inv_cov: NDArray[np.float64]):
+    """
+    Returns mahalanobis metric for matching. Works better with continuous covariates.
+
+    .. math::
+        D(x, y) = \\sqrt{(x-y)^T S^{-1} (x-y)} where S is the covariate matrix of
+        the whole distribution
```
JabobKrauskopf, line 74: Does this math equation work (in the context of formatting)? I've never seen it written like this but IDK

JabobKrauskopf (commented):
One question. The rest looks great :)

medmodels/matching/metrics.py
```diff
@@ -63,6 +65,43 @@ def exact_metric(vector1: NDArray[np.float64], vector2: NDArray[np.float64]) ->
     return np.inf
 
 
+def mahalanobis_metric(*vectors: NDArray[np.float64], inv_cov: NDArray[np.float64]):
```
JabobKrauskopf, line 68: oh and `*vectors` should probably be refactored into `vector_1` and `vector_2` or something similar

JabobKrauskopf (changes requested):

medmodels/matching/metrics.py
MarIniOnz: [Wikipedia article on Mahalanobis distance](https://en.wikipedia.org/wiki/Mahalanobis_distance)
image

And According to ChatGPT:
image

JabobKrauskopf (approved):
Amazeballs

medmodels/matching/metrics.py
```diff
@@ -63,6 +67,49 @@ def exact_metric(vector1: NDArray[np.float64], vector2: NDArray[np.float64]) ->
     return np.inf
 
 
+def mahalanobis_metric(
+    vector1: NDArray[np.float64],
+    vector2: NDArray[np.float64],
+    inv_cov: NDArray[np.float64],
```
LauraBoenchenLB, line 73: Based on the Mahalanobis formula, I would expect the covariance Matrix as input, so that the calculating of the inverse Covariance Matrix happens in the mahalanobis function and not before.

medmodels/matching/metrics.py
```diff
@@ -63,6 +67,49 @@ def exact_metric(vector1: NDArray[np.float64], vector2: NDArray[np.float64]) ->
...
+    repeated computation.
+
+    By matching without replacement the found paired item will be removed from the set,
+    hence,the covariance matrix of the whole distribution and its inverse need to be
+    recalculated for every entry. This can be time consuming for big data sets (esp.
+    with a big amount of features).
+    """
+
+    diff = vector1 - vector2
+    left = np.dot(diff, inv_cov)
```
LauraBoenchenLB, line 104: I researched the np.dot function a bit, because I was a bit confused by the missing transpose in this. I read that np.dot can have a bit of a confusing behaviour and using matrix multiplication operator @ and explicitly using the transpose could be more clear.

LauraBoenchenLB (changes requested):
Great PR, good testing coverage! Just two little comments of what I would expect looking at it from the math perspective.

medmodels/matching/metrics.py
MarIniOnz: Very good comment! Thanks. Now I know more about this!

MarIniOnz: From this comment, I realized that having a more clear mahalanobis function would be best. That is, including the difference between what happens in the three scenarios:
- 1-D (we do not use covariance really, but standard deviation)
- 2+-D (normal mahalanobis)
- Singular covariance matrix: in some cases, the inverse covariance cannot be computed since the matrix is singular, and an error should apper.

LauraBoenchenLB (dismissed):
Perfection! 😻

JabobKrauskopf (approved):
LGTM!

JabobKrauskopf (changes requested):
@MarIniOnz Can you update the branch? :)

JabobKrauskopf (approved):
LGTMLGTM

medmodels/matching/metrics.py
```diff
@@ -57,12 +59,56 @@ def exact_metric(vector1: NDArray[np.float64], vector2: NDArray[np.float64]) ->
...
+    """
+    Returns mahalanobis metric for matching. Works better with continuous covariates.
+
+    $$
+    D(x, y) = \\sqrt{(x-y)^T S^{-1} (x-y)} \\\\
+    \\text{where } S \\text{ is the covariance matrix of the whole distribution}
+    $$
+
+    The covariance matrix and its inverse are calculated at most once per item to be
+    paired, hence, they won't be included inside of the method in order to avoid the
```
LauraBoenchenLB, line 83: In that case it makes sense to keep it outside of the function, you're right :)

LauraBoenchenLB (approved):
Great PR, functions are clear and good testing coverage 👍

PR #106, opened 2024-06-17
feat: Pandas to polars, change of variable names and adding +1 neighbors.

MarIniOnz:
Changing the nearest neighbour function for better readability, adding more features to it (the posibility of retrieving more than 1 neighbor back) and changing the schema from pandas to polars dataframes.

### DIFF
--- a/medmodels/matching/algorithms/classic_distance_models.py
+++ b/medmodels/matching/algorithms/classic_distance_models.py
@@ -1,19 +1,21 @@
 from __future__ import annotations
 
-from typing import List, Optional, Union
+from typing import Optional
 
 import numpy as np
-import pandas as pd
+import polars as pl
 
 from medmodels.matching import metrics
+from medmodels.medrecord.types import MedRecordAttributeInputList
 
 
 def nearest_neighbor(
-    treated_set: pd.DataFrame,
-    control_set: pd.DataFrame,
-    metric: str,
-    covariates: Optional[Union[List[str], pd.Index[str]]] = None,
-) -> pd.DataFrame:
+    treated_set: pl.DataFrame,
+    control_set: pl.DataFrame,
+    metric: metrics.Metric,
+    number_of_neighbors: int = 1,
+    covariates: Optional[MedRecordAttributeInputList] = None,
+) -> pl.DataFrame:
     """
     Performs nearest neighbor matching between two dataframes using a specified metric.
     This method employs a greedy algorithm to pair elements from the treated set with
@@ -24,32 +26,32 @@ def nearest_neighbor(
     of matching. It supports optional specification of covariates for focused matching.
 
     Args:
-        treated_set (pd.DataFrame): DataFrame for which matches are sought.
-        control_set (pd.DataFrame): DataFrame from which matches are selected.
-        metric (str): Metric to measure closeness between units, e.g., "absolute",
-            "mahalanobis".
-        covariates (Optional[Union[List[str], pd.Index[str]]], optional): Covariates
+        treated_set (pl.DataFrame): DataFrame for which matches are sought.
+        control_set (pl.DataFrame): DataFrame from which matches are selected.
+        metric (metrics.Metric): Metric to measure closeness between units, e.g.,
+            "absolute", "mahalanobis". The metric must be available in the metrics
+            module.
+        number_of_neighbors (int, optional): Number of nearest neighbors to find for
+            each treated unit. Defaults to 1.
+        covariates (Optional[MedRecordAttributeInputList], optional): Covariates
             considered for matching. Defaults to all variables.
 
     Returns:
-        pd.DataFrame: Matched subset from the control set.
+        pl.DataFrame: Matched subset from the control set.
     """
-
-    columns = treated_set.columns
-
     if not covariates:
-        covariates = columns
+        covariates = treated_set.columns
 
-    treated_array = treated_set[covariates].to_numpy().astype(float)
-    control_array = control_set[covariates].to_numpy().astype(float)
+    treated_array = treated_set.select(covariates).to_numpy().astype(float)
+    control_array = control_set.select(covariates).to_numpy().astype(float)
     control_array_full = control_set.to_numpy()  # To keep all the information
-    matched_group = pd.DataFrame(columns=columns)
+    matched_control = []
 
     cov = np.array([])
     if metric == "mahalanobis":
         cov = np.cov(np.concatenate((treated_array, control_array)).T)
 
-    for element_ss in treated_array:
+    for treated_subject in treated_array:
         if metric == "mahalanobis":
             if cov.ndim == 0:
                 inv_cov = 1 / cov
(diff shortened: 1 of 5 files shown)


JabobKrauskopf:
There still are some variables that have a confusing (for me) name (mostly abbreviations). I'll refer to this video for naming "guidelines" (altough not everything is applicable).
For example `nn_indices` instead of `nearest_neighbor_indices` (i guess?)

medmodels/matching/algorithms/classic_distance_models.py
```diff
@@ -62,30 +63,35 @@ def nearest_neighbor(
...
-        )
-        # For the k:1 matching don't consider the chosen row any more.
-        control_array_full = np.delete(control_array_full, nn_index, 0)
-        control_array = np.delete(control_array, nn_index, 0)
+        for neighbor_index in neighbor_indices:
+            new_row = pl.DataFrame(
+                [control_array_full[neighbor_index]], schema=treated_set.columns
+            )
+            matched_group = (
+                new_row if matched_group.is_empty() else matched_group.vstack(new_row)
```
JabobKrauskopf, line 88: check `vstack` vs `concat`

medmodels/matching/algorithms/propensity_score.py
```diff
@@ -18,7 +19,7 @@ def calculate_propensity(
     treated_test: NDArray[Union[np.int64, np.float64]],
     control_test: NDArray[Union[np.int64, np.float64]],
     hyperparam: Optional[Dict[str, Any]] = None,
-    metric: str = "logit",
+    model: str = "logit",
```
JabobKrauskopf, line 22: Should also probably be a `Literal["logit", ...]`

medmodels/matching/algorithms/propensity_score.py
```diff
@@ -68,41 +69,47 @@ class for treated and control groups.
 
 
 def run_propensity_score(
-    treated_set: pd.DataFrame,
-    control_set: pd.DataFrame,
+    treated_set: pl.DataFrame,
+    control_set: pl.DataFrame,
     model: str = "logit",
```
JabobKrauskopf, line 74: Same as above

medmodels/matching/algorithms/propensity_score.py
```diff
@@ -68,41 +69,47 @@ class for treated and control groups.
...
+    metric: str = "absolute",
```
JabobKrauskopf, line 75: Same as above

medmodels/matching/algorithms/propensity_score.py
```diff
@@ -68,41 +69,47 @@ class for treated and control groups.
...
         model (str, optional): Classification algorithm for predicting probabilities.
             Options include "logit", "dec_tree", "forest".
+        metric (str, optional): Metric for matching. Options include "absolute",
+            "mahalanobis", "exact".
+        num_neighbors (int): Number of nearest neighbors to find for each treated unit.
+            Defaults to 1.
         hyperparam (Optional[Any], optional): Hyperparameters for model tuning.
             Increases computation time if set.
-        covariates (Optional[Union[List[str], pd.Index[str]]], optional): Features for
+        covariates (Optional[Union[List[str], pl.Index]], optional): Features for
```
JabobKrauskopf, line 97: type should also probably be changed to `MedRecordAttributeInputList`

JabobKrauskopf (changes requested):
Noice! Some minor changes requested :)

medmodels/matching/algorithms/propensity_score.py
JabobKrauskopf: I also don't think `pl.Index` is a thing :D

LauraBoenchenLB (approved):
Great PR, typing looks really good and cohesive :)

pyproject.toml
```diff
@@ -9,6 +9,7 @@ dependencies = [
     "pandas==2.2.2",
     "polars[pandas]==0.20.31",
     "scikit-learn==1.5.0",
+    "typing-extensions==4.12.2",
```
JabobKrauskopf, line 12: `typing_extensions` is a package that is automatically available for python version < 3.10. This is not the way to solve the typing problem

medmodels/matching/metrics.py
```diff
@@ -1,9 +1,11 @@
 """Metrics for comparing vectors in the context of matching classes."""
 
 import math
+from typing import Literal
 
 import numpy as np
 from numpy.typing import NDArray
+from typing_extensions import TypeAlias
```
JabobKrauskopf, line 8: See https://github.com/limebit/medmodels/blob/main/medmodels/medrecord/types.py#L8-L14 on how to import this correctly

medmodels/matching/algorithms/propensity_score.py
```diff
@@ -1,24 +1,27 @@
...
+from typing import Any, Dict, Literal, Optional, Tuple, Union
 
 import numpy as np
-import pandas as pd
+import polars as pl
 from numpy.typing import NDArray
 from sklearn.ensemble import RandomForestClassifier
 from sklearn.linear_model import LogisticRegression
 from sklearn.tree import DecisionTreeClassifier
+from typing_extensions import TypeAlias
```
JabobKrauskopf, line 11: See https://github.com/limebit/medmodels/blob/main/medmodels/medrecord/types.py#L8-L14 on how to import this correctly

medmodels/matching/algorithms/propensity_score.py
```diff
@@ -114,24 +122,31 @@ def run_propensity_score(
...
+    matched_control = matched_control.drop("prop_score")
+    treated_set = treated_set.drop("prop_score")
+    control_set = control_set.drop("prop_score")
 
     return matched_control
 
 
+Model: TypeAlias = Literal["logit", "dec_tree", "forest"]
+
+
```
JabobKrauskopf, line 149: Static definitions and type definitions (`PROP_MODEL` and `Model`) should go to the top of the module

medmodels/matching/algorithms/classic_distance_models.py
```diff
@@ -1,19 +1,21 @@
 from __future__ import annotations
```
JabobKrauskopf, line 1: This should not be necessary for this module

JabobKrauskopf: `from __future__ import annotations` should be used sparingly

medmodels/matching/algorithms/propensity_score.py
```diff
@@ -1,24 +1,27 @@
 from __future__ import annotations
```
JabobKrauskopf, line 1: This should not be necessary for this module

medmodels/matching/algorithms/classic_distance_models.py
```diff
@@ -62,30 +64,35 @@ def nearest_neighbor(
...
-        control_array = np.delete(control_array, nn_index, 0)
+            # For the k:1 matching don't consider the chosen row any more.
+            control_array_full = np.delete(control_array_full, neighbor_index, 0)
+            control_array = np.delete(control_array, neighbor_index, 0)
 
-    return matched_group.reset_index(drop=True)
+    return pl.concat(matched_control, how="vertical")
 
 
 ALGORITHMS = {"nearest neighbor": nearest_neighbor}
```
JabobKrauskopf, line 98: Static definition to the top of the module as per one of the other comments

medmodels/matching/metrics.py
```diff
@@ -109,6 +111,8 @@ def mahalanobis_metric(
     return float(math.sqrt(distance_squared))
 
 
+Metric: TypeAlias = Literal["absolute", "exact", "mahalanobis"]
```
JabobKrauskopf, line 114: Static variables and TypeAliases to the top of the module as per one of the other comments

JabobKrauskopf (changes requested):
Some minor typing change requests (including some code that was not changed in this PR but would fit into this PR)

pyproject.toml
LauraBoenchenLB: >3.10?

JabobKrauskopf: nope versions < 3.10
After that it is available in the `typing package`

JabobKrauskopf: @LauraBoenchenLB ey lösch doch dein kommentar nich einfach. Pff

medmodels/matching/algorithms/classic_distance_models.py
JabobKrauskopf: i was wrong

medmodels/matching/algorithms/classic_distance_models.py
JabobKrauskopf: This does not work in this specific case because python is hilariously stupid

medmodels/matching/metrics.py
JabobKrauskopf: the static part also does not work in this case

PR #107, opened 2024-06-18
feat: Add undirected versions of edges_connecting and neighbors

JabobKrauskopf:


crates/medmodels-core/src/medrecord/graph/mod.rs
```diff
@@ -313,18 +316,34 @@ impl Graph {
 
     pub fn edges_connecting<'a>(
         &'a self,
-        source_node_index: Vec<&'a NodeIndex>,
+        source_node_indices: Vec<&'a NodeIndex>,
         target_node_index: Vec<&'a NodeIndex>,
```
MarIniOnz, line 320: Why change source_node_index naming and not also target_node_index?

MarIniOnz (approved):
Very nice feature added! Tests reviewed and everything. Just a small naming comment

crates/medmodels-core/src/medrecord/graph/mod.rs
JabobKrauskopf: good catch, thank you :)
Fixed

PR #122, opened 2024-07-08
chore: Adjust dependencies and increase version

JabobKrauskopf:


MarIniOnz (commented):
We are on the last run! Publication is coming!

MarIniOnz (approved):

PR #124, opened 2024-07-09
test: Add tests for MedRecord schema and builder

JabobKrauskopf:


crates/medmodels-core/src/medrecord/group_mapping.rs
```diff
@@ -313,15 +313,22 @@ mod test {
         assert_eq!(1, group_mapping.group_count());
 
         group_mapping
-            .add_group("1".into(), Some(vec!["0".into(), "1".into()]), None)
+            .add_group(
```
MarIniOnz, line 316: Missing the test on adding a node that does not exist to a group. Is it there already and I have not found it?

MarIniOnz (commented):
Great! New tests added, Amazing PRR

crates/medmodels-core/src/medrecord/group_mapping.rs
JabobKrauskopf: The Group Mapping itself has no knowledge if a node exists or not. the medrecord does. there is a test for this case in the medrecord :)

crates/medmodels-core/src/medrecord/group_mapping.rs
```diff
@@ -506,6 +638,17 @@ mod test {
         assert_eq!(1, group_mapping.groups_of_node(&"0".into()).count());
     }
 
+    #[test]
+    fn test_groups_of_edge() {
```
MarIniOnz, line 642: Would we need an invalid_groups_of_edge test or invalid_groups_of_node?
Or is there a case where the count of these groups should be 0?

crates/medmodels-core/src/medrecord/querying/selection.rs
```diff
@@ -939,7 +939,11 @@ mod test {
 
     #[test]
     fn test_select_edges_edge() {
-        let medrecord = create_medrecord();
+        let mut medrecord = create_medrecord();
+
+        medrecord
+            .add_group("test".into(), None, Some(vec![0]))
+            .unwrap();
```
MarIniOnz, line 946: Invalid test for this needed?

medmodels/medrecord/tests/test_querying.py
```diff
@@ -609,6 +611,12 @@ def test_select_edges_edge(self):
             sorted(medrecord.select_edges(edge().connected("1"))),
         )
 
+        # Edge in group
```
MarIniOnz, line 614: Do we get an invalid in_group test?

MarIniOnz (changes requested):
Amazing PR! 
Just some minor questions. Might be actually all okay already.

crates/medmodels-core/src/medrecord/group_mapping.rs
JabobKrauskopf: the `invalid_` tests are tests for when an error gets returned. The `groups_of_edge` and `groups_of_node` always return an iterator (which might be empty) so there is no invalid way of calling this function.
Checking for an iterator of size zero checks the exact same code path that is already being checked, so this is not necessary imo.

crates/medmodels-core/src/medrecord/querying/selection.rs
JabobKrauskopf: adding the group here is not part of the actual test, only the setup.

medmodels/medrecord/tests/test_querying.py
JabobKrauskopf: any query that is syntactically correct (which is checked by pyright) is a valid query => there can be no test for an invalid query because they do not exist

MarIniOnz (approved):
image

PR #129, opened 2024-07-10
refactor: full treatment effect report now returns typed dictionary

JabobKrauskopf:
There was no way of knowing the keys of the dictionary returned by `full_report` (except for checking the function itself). I replaced it with a typed dictionary. The keys are now statically typed and are available in auto completion


MarIniOnz (approved):
Amazing! Easy and to the point!
DALL·E 2024-07-10 14 48 10 - A group of seals on a beach, looking at the camera and giving thumbs up  The seals are anthropomorphized, with friendly and happy expressions  The bea

Issue #140, opened 2024-07-17
Reconfigure Medmodels folder


MarIniOnz:
Matching folder should belong within Treatment Effect Estimation. If any functions could be common to other modules, set up a new module known as "utils".

PR #143, opened 2024-07-17
chore: move files and delete the ones that need refactoring (issues 141, 142)

MarIniOnz:
Moved files into treatment_effect folder.

No common utils function was created. In case we find some functions can and should be shared, we will take care of that when it applies.

Deleted covariates_preprocessing.py and evaluation.py together with their tests and set issues to refactor them accordingly.
Issue: [#140]( https://github.com/limebit/medmodels/issues/140 )


MarIniOnz:
Wrong branch naming.

PR #159, opened 2024-08-08
feat: make schema accessible from python and add AttributeType

JabobKrauskopf:


crates/medmodels-core/src/medrecord/schema.rs
```diff
@@ -8,7 +8,47 @@ use crate::{
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, HashSet};
 
-type AttributeSchema = HashMap<MedRecordAttribute, DataType>;
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub enum AttributeType {
+    Categorical,
+    Regression,
```
LauraBoenchenLB, line 14: Regression as a name for the AttributeType is not the best choice. Maybe Continous or Quantitative would be better to fit with Categorical. Regression is a statistical technique where the result is that type for variable, but you would not call an variable type regression.

LauraBoenchenLB (changes requested):
Looks great, very excited for the possibilities this will bring! One small comment regarding the naming.

PR #183, opened 2024-08-20
docs: update developer guide

JabobKrauskopf:


docs/developer_guide/setup.md
```diff
@@ -0,0 +1,25 @@
...
+MedModels leverages a combination of Python and Rust code. To contribute effectively, you'll need to set up a development environment that supports both languages.
+
+**Requirements:**
+
+- Python (3.9, 3.10, 3.11 or 3.12)
+- Rust compiler (follow instructions from [https://www.rust-lang.org/tools/install](https://www.rust-lang.org/tools/install))
+
+**Using the Makefile:**
+
+MedModels utilizes a `Makefile` to manage development tasks. Here's a breakdown of the available commands and their functionalities:
```
MarIniOnz, line 12: We are missing this information that was included before. Do we want to get rid of it? We do not show how Makefile commands are run.

**Using the Makefile with your IDE:**

The provided `Makefile` can be integrated with most IDEs. Consult your IDE's documentation for instructions on adding custom build tasks based on Makefiles. This allows you to easily run commands like `make install-dev` or `make test` directly from within your IDE.

By following these steps and leveraging the `Makefile`, you'll have a development environment ready to contribute to the MedModels codebase!

docs/developer_guide/internal-workflow.md
```diff
@@ -0,0 +1,112 @@
...
+An epic branch is created for each epic issue. Completed sub-tasks are merged into this branch via pull requests.
+
+#### Epic Task Pull Requests
+
+- May include failing tests (with explanations)
+- Must still adhere to linting and formatting rules and follow the [pull request guidelines](./pull-request.md)
+
+#### Epic Pull Request
+
+The final pull request merges the epic branch into the main branch and must include links to all sub-task pull requests.
```
MarIniOnz, line 110: Maybe state that all tasks should be finished before merging an epic branch?

docs/developer_guide/internal-workflow.md
```diff
@@ -0,0 +1,112 @@
...
+  "python.terminal.activateEnvInCurrentTerminal": true,
+  "autoDocstring.docstringFormat": "google",
+  "python.analysis.typeCheckingMode": "strict",
+  "[python]": {
+    "editor.defaultFormatter": "charliermarsh.ruff"
+  }
+}
+```
+
+## Issue Management
```
MarIniOnz, line 65:
```suggestion
## Issues Management
```

MarIniOnz (changes requested):
Very noice PR. Minor changes suggested/ needing some clarification

docs/developer_guide/setup.md
JabobKrauskopf: I removed this part because I didn't find it necessary to include it in the guide in the first place but I can add it back if you find it helpful.

Makefiles are practically always run using `make ..`, adding it to the IDE is just a convenience that is not necessary. IMO if a user wants to do it, go ahead but I wouldn't suggest it. That feels out of place in the medmodels guide

docs/developer_guide/internal-workflow.md
JabobKrauskopf: singular is correct here imo (and apparently chatgpts opinion). Similar to how `Issue Types` is `Issue Types` and not `Issues Types`

MarIniOnz (approved):
image

philippgrosser (approved):

PR #190, opened 2024-09-03
feat: Complete Synthesizer, SynthesizerModel and SynthesizerBuilder

MarIniOnz:
Adding the abstract classes for Synthesizer, SynthesizerBuilder and SynthesizerModel.

Eliminating associated interfaces.


medmodels/data_synthesis/builder.py
```diff
@@ -0,0 +1,15 @@
+"""Builder for data Synthesizers."""
+
+from __future__ import annotations
+
+from medmodels.data_synthesis.mtgan.builder import MTGANBuilder
+
+
+class SynthesizerBuilder:
+    def from_mtgan(self) -> MTGANBuilder:
+        """Creates a MTGANBuilder instance from a SynthesizerBuilder instance.
```
JabobKrauskopf, line 10:
```suggestion
        """Creates an MTGANBuilder instance.
```

medmodels/data_synthesis/synthesizer.py
```diff
@@ -0,0 +1,124 @@
...
+
+if TYPE_CHECKING:
+    from medmodels.data_synthesis.builder import SynthesizerBuilder
+
+
+class Synthesizer(nn.Module, metaclass=ABCMeta):
+    """Synthesizer is an abstract class that serves as a blueprint for synthesizers.
+
+    It ensures the correct instantiation, training, and persistence of models.
+    """
```
JabobKrauskopf, line 23: Sounds very chatgpt. Could also be way shorter

medmodels/data_synthesis/synthesizer.py
```diff
@@ -0,0 +1,124 @@
...
+        checkpoint directory if it is provided. It will take the checkpoint paths
+        where the model is saved as keyword arguments to start fitting from them.
+
+        It is an abstract method that should be implemented in the child class.
+
+        Args:
+            medrecord (MedRecord): MedRecord object.
+            save_directory (Optional[Path]): Directory to save the model to.
+            **kwargs (Dict[str, Path]): Additional keyword arguments.
+
```
JabobKrauskopf, line 101: needs to be updated

JabobKrauskopf: also explain the kwargs

medmodels/data_synthesis/synthesizer.py
```diff
@@ -0,0 +1,124 @@
...
+        from medmodels.data_synthesis.builder import SynthesizerBuilder
+
+        return SynthesizerBuilder()
+
+    @abstractmethod
+    def fit(
+        self,
+        medrecord: MedRecord,
+        checkpoint_directory: Optional[Path],
+        **kwargs: Dict[str, Any],
```
JabobKrauskopf, line 62: wrong type

medmodels/data_synthesis/synthesizer.py
```diff
@@ -0,0 +1,124 @@
...
+
+        This method is meant to fit the synthesizer on the given data on scratch (no
+        pre-trained model is used).
+
+        It is an abstract method that should be implemented in the child class.
+
+        Args:
+            medrecord (MedRecord): MedRecord object.
+            checkpoint_directory (Optional[Path]): Directory to save the model to.
+            **kwargs (Dict[str, Any]): Additional keyword arguments.
```
JabobKrauskopf, line 74: explain

medmodels/data_synthesis/synthesizer.py
```diff
@@ -0,0 +1,124 @@
...
+            **kwargs (Dict[str, Path]): Additional keyword arguments.
+
+        Returns:
+            Synthesizer: Trained synthesizer.
+        """
+        pass
+
+    @abstractmethod
+    def load_model(
+        self, medrecord: MedRecord, path: Path, **kwargs: Dict[str, Any]
```
JabobKrauskopf, line 109: wrong `**kwargs` type and missing explanation

medmodels/data_synthesis/synthesizer.py
```diff
@@ -0,0 +1,124 @@
...
+            postprocessor (Type[nn.Module]): Type of postprocessor module.
+        """
+        super(Synthesizer, self).__init__()
+        self._preprocessor = preprocessor
+        self._postprocessor = postprocessor
+        self.device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+        self.to(self.device)
+
+    @classmethod
+    def builder(cls) -> SynthesizerBuilder:
```
JabobKrauskopf, line 47:
```suggestion
    @staticmethod
    def builder() -> SynthesizerBuilder:
```

medmodels/data_synthesis/synthesizer.py
```diff
@@ -0,0 +1,124 @@
...
+        """Initializes the synthesizer with the required components.
+
+        Args:
+            preprocessor (Type[nn.Module]): Type of preprocessor module.
+            postprocessor (Type[nn.Module]): Type of postprocessor module.
+        """
+        super(Synthesizer, self).__init__()
+        self._preprocessor = preprocessor
+        self._postprocessor = postprocessor
+        self.device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
```
JabobKrauskopf, line 43: the device should be selectable, not automatically use the gpu if the user might not want to

medmodels/data_synthesis/synthesizer_model.py
```diff
@@ -0,0 +1,84 @@
...
+from medmodels import MedRecord
+from medmodels.data_synthesis.synthesizer import Synthesizer
+
+
+class SynthesizerModel(nn.Module, metaclass=ABCMeta):
+    """SynthesizerModel is an abstract class that serves as a blueprint for synthesizer models.
+
+    This class is instantiated once a Synthesizer has been trained on real data. It is
+    separated from the Synthesizer class to allow for the generation of synthetic data
+    without incurring into side effects."""
```
JabobKrauskopf, line 22: also sounds very chatgpt.

`allow for the generation of synthetic data without incurring into side effects.`
this sentence does not make sense and there still are side effects, just more expected ones

medmodels/data_synthesis/synthesizer_model.py
```diff
@@ -0,0 +1,84 @@
...
+        self,
+        synthesizer: Synthesizer,
+        number_of_samples: int,
+        device: torch.device,
+        **kwargs: Dict[str, Any],
+    ) -> None:
+        """Initializes the SynthesizerModel with the required components.
+
+        Args:
+            synthesizer (Synthesizer): A trained synthesizer model.
```
JabobKrauskopf, line 37: is this correct?

medmodels/data_synthesis/synthesizer_model.py
```diff
@@ -0,0 +1,84 @@
...
+
+    number_of_samples: int
+    device: torch.device
+
+    def __init__(
+        self,
+        synthesizer: Synthesizer,
+        number_of_samples: int,
+        device: torch.device,
+        **kwargs: Dict[str, Any],
```
JabobKrauskopf, line 32: explain the kwargs

medmodels/data_synthesis/synthesizer_model.py
```diff
@@ -0,0 +1,84 @@
...
+        """Saves the trained synthesizer model to a file.
+
+        Args:
+            path (Path): Path to save the model to.
+        """
+        pass
+
+    @abstractmethod
+    def generate_synthetic_data(self) -> MedRecord:
+        """Generates synthetic data using the trained synthesizer model.#
```
JabobKrauskopf, line 67: remove `#`

also go into detail that this MedRecord is not postprocessed yet and cannot be really used

medmodels/data_synthesis/synthesizer_model.py
```diff
@@ -0,0 +1,84 @@
...
+        """Generates synthetic data using the trained synthesizer model.
+
+        Returns:
+            Union[torch.Tensor, sparse.COO]: Synthetic data generated by the model.
+        """
+        pass
+
+    @abstractmethod
+    def save_model(self, path: Path) -> None:
+        """Saves the trained synthesizer model to a file.
```
JabobKrauskopf, line 58: is it always trained already?

JabobKrauskopf (changes requested):

medmodels/data_synthesis/synthesizer.py
```diff
@@ -0,0 +1,124 @@
...
+    @abstractmethod
+    def fit(
+        self,
+        medrecord: MedRecord,
+        checkpoint_directory: Optional[Path],
+        **kwargs: Dict[str, Any],
+    ) -> SynthesizerModel:
+        """Trains the synthesizer on the given data and saves the model if a directory is provided.
+
+        This method is meant to fit the synthesizer on the given data on scratch (no
```
LauraBoenchenLB, line 66: from scratch

medmodels/data_synthesis/synthesizer.py
```diff
@@ -0,0 +1,124 @@
...
+
+        It is an abstract method that should be implemented in the child class.
+
+        Args:
+            medrecord (MedRecord): MedRecord object.
+            save_directory (Optional[Path]): Directory to save the model to.
+            **kwargs (Dict[str, Path]): Additional keyword arguments.
+
+        Returns:
+            Synthesizer: Trained synthesizer.
```
LauraBoenchenLB, line 103: Returns SynthesizerModel

medmodels/data_synthesis/synthesizer.py
```diff
@@ -0,0 +1,124 @@
...
+        Returns:
+            SynthesizerModel: Trained synthesizer.
+        """
+        pass
+
+    @abstractmethod
+    def fit_from(
+        self,
+        medrecord: MedRecord,
+        checkpoint_directory: Optional[Path],
```
LauraBoenchenLB, line 85: Is the checkpoint_directory really optional? Where does the pre-trained model come from or is it in the kwargs?

medmodels/data_synthesis/synthesizer.py
```diff
@@ -0,0 +1,124 @@
...
+        self, medrecord: MedRecord, path: Path, **kwargs: Dict[str, Any]
+    ) -> SynthesizerModel:
+        """Loads the model from a specified path.
+
+        This method is meant to load the model from the given path. It is an abstract
+        method that should be implemented in the child class.
+
+        Args:
+            medrecord (MedRecord): MedRecord object.
+            path (Path): Path to load the model parameters from.
```
LauraBoenchenLB, line 118: Loads the model parameters or the actual model?

medmodels/data_synthesis/synthesizer_model.py
```diff
@@ -0,0 +1,84 @@
...
+    def generate_synthetic_data(self) -> MedRecord:
+        """Generates synthetic data using the trained synthesizer model.#
+
+        Returns:
+            MedRecord: Synthetic data generated by the model.
+        """
+        pass
+
+    @abstractmethod
+    def postprocess(self, synthetic_data: Union[MedRecord, sparse.COO]) -> MedRecord:
```
LauraBoenchenLB, line 75: I'm a bit confused by the Union here, is it just to make sure that all the different postprocess methods are included? Here it looks like the `generate_synthetic_data` function is only returning a MedRecord, shouldn't the input for the postprocessing also be that then? Or can it also be postprocessed directly from the `forward` output?

LauraBoenchenLB (changes requested):

medmodels/data_synthesis/synthesizer.py
MarIniOnz: You are right, we need a checkpoint directory to which we will save the result of this training, and another checkpoint_path from which we will load the pre-trained model. I am not convinced with the naming though. Would you have suggestions @JabobKrauskopf ?

JabobKrauskopf: maybe `checkpoint_directory` and `model_path`?

medmodels/data_synthesis/synthesizer_model.py
MarIniOnz: Yes! The SynthesizerModel is always a trained one. That is the idea of separating them.

medmodels/data_synthesis/synthesizer_model.py
MarIniOnz: The idea is that the generate_synthetic_data method combines both the forward and postprocessing methods. Thus, the postprocess would take in the forward method. If a better architecture could be devised, please let me know! :D

medmodels/data_synthesis/synthesizer.py
```diff
@@ -0,0 +1,132 @@
...
+
+    _preprocessor: Type[nn.Module]
+    _postprocessor: Type[nn.Module]
+    device: torch.device
+
+    def __init__(
+        self,
+        preprocessor: Type[nn.Module],
+        postprocessor: Type[nn.Module],
+        device: Literal["cpu", "cuda"] = "cuda",
```
JabobKrauskopf, line 30: there are many more types other than `cpu` or `cupa`, including e.g. `mps` for macbooks.

I'd probably suggest
```suggestion
        device: torch.device = torch.device("cpu"),
```

medmodels/data_synthesis/synthesizer.py
```diff
@@ -0,0 +1,132 @@
...
+        from medmodels.data_synthesis.builder import SynthesizerBuilder
+
+        return SynthesizerBuilder()
+
+    @abstractmethod
+    def fit(
+        self,
+        medrecord: MedRecord,
+        checkpoint_directory: Optional[Path],
+        **kwargs: Any,
```
JabobKrauskopf, line 64: if you really want to pass anything as kwargs then `Dict[str, Any]` was correct. I though that you once said that these kwargs were for extra paths or sth, then `Dict[str, Path]` but I probably remember wrong.

Can you give me an example of which additional kwargs might be used in the "child method"? Because from what i udnerstand now, this would violate the concept of subclasses. The entire point of ABC ist to define which methods exist, including their parameters (all of them, subclasses should not have different parameters as this defeats the purpose of using ABC)

medmodels/data_synthesis/synthesizer.py
```diff
@@ -0,0 +1,132 @@
...
+    ) -> None:
+        """Initializes the synthesizer with the required components.
+
+        Args:
+            preprocessor (Type[nn.Module]): Type of preprocessor module.
+            postprocessor (Type[nn.Module]): Type of postprocessor module.
+            device (Literal["cpu", "cuda"], optional): Device to use for the model. Defaults
+                to "cuda".
+        """
+        super(Synthesizer, self).__init__()
```
JabobKrauskopf, line 40:
```suggestion
        super().__init__()
```

should be enough. The other syntax is python 2.0 afaik

medmodels/data_synthesis/synthesizer_model.py
```diff
@@ -0,0 +1,87 @@
...
+        """Initializes the SynthesizerModel with the required components.
+
+        Args:
+            synthesizer (Synthesizer): An untrained Synthesizer Model.
+            number_of_samples (int): Number of samples to generate.
+            device (torch.device): Device to use for the model.
+            **kwargs (Any): Additional keyword arguments. These arguments should be
+                specified in the child class.
+        """
+        super(SynthesizerModel, self).__init__()
```
JabobKrauskopf, line 42:
```suggestion
        super().__init__()
```

JabobKrauskopf (changes requested):
image

medmodels/data_synthesis/synthesizer.py
JabobKrauskopf: either all of these attributes are used in the abstract method definition (and therefore any future synthesizer implementations) or we need to rethink the whole ABC and subclass structure

MarIniOnz: It is something very specific to MTGAN, do not know whether it makes sense to add it there. Although it could be added for now and then rethought if more synthesizers are subsequently added?

JabobKrauskopf: I'd prefer if we chose the correct solution now tbh.

If the attributes of the actual implementations of the functions are not the same, they cannot be defined as abstract functions. That would defeat the purpose of ABC.
If this is the case, we should remove those abstract definitions.

MarIniOnz: Then, removing all abstract methods that could not be shared? Or removing the abstract classes themselves?

JabobKrauskopf: remove all abstract methods (where the actual implementations require different attributes) and if all abstract methods are removed, also remove the abstract declaration of the class (the `metaclass=ABCMeta` part)

medmodels/data_synthesis/synthesizer_model.py
```diff
@@ -0,0 +1,87 @@
...
+
+    number_of_samples: int
+    device: torch.device
+
+    def __init__(
+        self,
+        synthesizer: Synthesizer,
+        number_of_samples: int,
+        device: torch.device,
+        **kwargs: Any,
```
JabobKrauskopf, line 31: this is not an abstract method, so any unused arguments should be removed

```suggestion
    def __init__(
        self,
        number_of_samples: int,
        device: torch.device,
```

However this might be an actual abstract method but then you'd have to remove the `**kwargs` (same discussion as in synthesizer)

JabobKrauskopf (changes requested):
almost done :D

medmodels/data_synthesis/mtgan/mtgan_model.pyi
```diff
@@ -27,3 +29,14 @@ class MTGANModel(SynthesizerModel):
...
+    def save_model(
+        self,
+        path: Path,
+    ) -> None: ...
+    def generate_synthetic_data(
+        self,
+    ) -> MedRecord: ...
+    def postprocess(
+        self,
+        synthetic_data: sparse.COO,
```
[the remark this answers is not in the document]
JabobKrauskopf: input variables need to be less specific than its abstract counterpart, so `Union[MedRecord, sparse.COO, ...]` would be allowed here, but not less.

return values of the abstract method implementation can be more specific than its abstract counterpart, see FYI comment above

JabobKrauskopf (changes requested):

LauraBoenchenLB (approved):
It is looking great! Nice PR, no comments, just approval 🦭

PR #197, opened 2024-09-11
feat: interface for comparer 

LauraBoenchenLB:


medmodels/statistic_evaluations/comparer/builder.pyi
```diff
@@ -0,0 +1,34 @@
+from __future__ import annotations
```
MarIniOnz, line 1: This line is not needed for the interface.
```suggestion

```

PR #200, opened 2024-09-12
feat: Add MTGAN preprocessor class

MarIniOnz:
Adding the _**MTGANPreprocessor**_ class, which will enable to add the necessary attributes to the _**MedRecord**_ to use it in **_MTGAN_**. It also removes nodes/edges if they do not fulfil the criteria given by the **_PreprocessingHyperparameters_**.

Example of usage
```python

medrecord = create_medrecord()
preprocessor = MTGANPreprocessor(hyperparameters)
preprocessed_medrecord, index_to_concept_dictionary, preprocessing_attributes = preprocessor.preprocess(medrecord)

```


medmodels/data_synthesis/mtgan/modules/preprocessor.py
```diff
@@ -0,0 +1,595 @@
...
+        """
+        patient_edges = medrecord.edges_connecting(
+            patient_index,
+            sorted(medrecord.nodes_in_group(self.concepts_group)),
+            directed=False,
+        )
+        time_windows = [
+            medrecord.edge[e, absolute_time_attribute] for e in patient_edges
+        ]
+        if not min_codes_per_window:
```
LauraBoenchenLB, line 329: This will never be true if `min_codes_per_window` is of int type

medmodels/data_synthesis/mtgan/modules/preprocessor.py
```diff
@@ -0,0 +1,595 @@
...
+    ) -> Tuple[MedRecord, PreprocessingAttributes]:
+        """Finds the relative time windows for each patient and removes the time windows with less than min_codes_per_window codes.
+
+        For instance, consider a single patient, 'p1', with admission dates as follows:
+        'p1': ['2021-01-01', '2021-01-15', '2021-01-25']
+
+        If the `time_interval_days` is 7 days, the method will compute time window as
+        the floor division of the difference in days from the first admission and the
+        `time_interval_days`. So, the resulting ABSOLUTE time windows will be:
+        'p1': [0, 2, 2]
```
LauraBoenchenLB, line 440: wouldn't it be `'p1': [0, 2, 3]`?

medmodels/data_synthesis/mtgan/modules/preprocessor.py
```diff
@@ -0,0 +1,595 @@
...
+                time = medrecord.edge[single_edge, self.time_attribute]
+                if not isinstance(time, datetime.datetime):
+                    raise ValueError(
+                        f"Time attribute needs to be a datetime object, but got {type(time)}"
+                    )
+                medrecord.edge[single_edge, absolute_time_window_attribute] = floor(
+                    (time - first_admission).days // time_interval_days
+                )
+                medrecord.edge[single_edge, concept_index_attribute] = (
+                    medrecord.select_nodes(
```
LauraBoenchenLB, line 491: So we're only keeping one concept per time window? How do we decide which one?

medmodels/data_synthesis/mtgan/modules/preprocessor.py
```diff
@@ -0,0 +1,595 @@
...
+        # TODO: copy of a MedRecord instead of modifying the original one once it is implemented
+        concept_index_attribute = self._get_attribute_name(medrecord, "concept_index")
+        medrecord, _ = self._remove_uncommon_concepts(
+            medrecord,
+            minimum_number_ocurrences=self.hyperparameters[
+                "minimum_occurrences_concept"
+            ],
+            concept_index_attribute=concept_index_attribute,
+        )
+        medrecord = self._sample_patients(
```
LauraBoenchenLB, line 569: Afterwards we remove patients again, so we will probably never get the number of patients that is stated in the Preprocessor

medmodels/data_synthesis/mtgan/modules/tests/test_preprocessor.py
```diff
@@ -0,0 +1,548 @@
+"""Tests for the MTGAN Preprocessor class in the data_synthesis module."""
```
LauraBoenchenLB, line 1:
```suggestion
"""Tests for the MTGAN Preprocessor class in the data synthesis module."""
```

medmodels/data_synthesis/mtgan/modules/tests/test_preprocessor.py
```diff
@@ -0,0 +1,548 @@
...
+
+        # If already exists, and the one with a suffix also does, adds 1 to the number
+        attribute_name = "age"
+        medrecord.add_node("new_patient", {"age_1": 30})
+        self.assertEqual(
+            self.preprocessor._get_attribute_name(medrecord, attribute_name), "age_2"
+        )
+
+    def test_remove_uncommon_concepts(self):
+        """Tests the test_remove_uncommon_concepts method of the MTGANPreprocessor class."""
```
LauraBoenchenLB, line 264:
```suggestion
        """Tests the remove_uncommon_concepts method of the MTGANPreprocessor class."""
```

medmodels/data_synthesis/mtgan/modules/tests/test_preprocessor.py
```diff
@@ -0,0 +1,548 @@
...
+        medrecord, first_admission_attribute = self.preprocessor._find_first_admission(
+            medrecord
+        )
+        self.assertEqual(
+            medrecord.node["P1", first_admission_attribute],
+            datetime.datetime(2000, 1, 1),
+        )
+
+    def test_preprocess(self):
+        """Tests the find_relative_times method of the MTGANPreprocessor class."""
```
LauraBoenchenLB, line 389:
```suggestion
        """Tests the preprocess method of the MTGANPreprocessor class."""
```

LauraBoenchenLB (changes requested):
Great Start to this very complex topic! A few things are still unclear and need to be addressed, but amazing what you have done so far. Also needs to be updated once we have the new query engine for even more speed and efficiency ⚡

medmodels/data_synthesis/mtgan/modules/preprocessor.py
```diff
@@ -0,0 +1,595 @@
...
+
+        if not return_dictionary:
+            return medrecord, {}
+
+        index_to_concept_dictionary = {}
+        for concept_index, concept in enumerate(
+            sorted(medrecord.nodes_in_group(self.concepts_group))
+        ):
+            index_to_concept_dictionary[concept_index] = concept
+            medrecord.node[concept, concept_index_attribute] = concept_index
```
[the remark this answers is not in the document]
MarIniOnz: Yeah, I think I should separate both parts into different functions. One that removes uncommon concepts and one and gets the index to concept dictionary and assigns it to the concepts.

medmodels/data_synthesis/mtgan/modules/preprocessor.py
MarIniOnz: you are right! Nicely found! Good understanding!

medmodels/data_synthesis/mtgan/modules/preprocessor.py
MarIniOnz: No, we are keeping all of them. This is done inside a loop over all edges: Thus all concepts! Let me know if more clarification is needed.

medmodels/data_synthesis/mtgan/modules/preprocessor.py
```diff
@@ -0,0 +1,590 @@
...
+    minimum_concepts_per_window: int
+    number_of_sampled_patients: int
+
+
+class PreprocessingAttributes(TypedDict):
+    first_admission_attribute: MedRecordAttribute
+    time_window_attribute: MedRecordAttribute
+    concept_index_attribute: MedRecordAttribute
+    number_of_windows_attribute: MedRecordAttribute
+    absolute_time_window_attribute: MedRecordAttribute
```
JabobKrauskopf, line 65: missing attribute from the interface file. Intentional?

medmodels/data_synthesis/mtgan/modules/preprocessor.py
```diff
@@ -0,0 +1,590 @@
...
+    hyperparameters: PreprocessingHyperparameters
+    seed: int
+
+    def __init__(
+        self,
+        hyperparameters: PreprocessingHyperparameters,
+        patients_group: Group = "patients",
+        concepts_group: Group = "concepts",
+        time_attribute: MedRecordAttribute = "time",
+        seed: int = 42,
```
JabobKrauskopf, line 83: Hitchhikers guide to the Galaxy reference?

seed should be optional imo.
```suggestion
        seed: Optional[int] = None,
```

medmodels/data_synthesis/mtgan/modules/preprocessor.py
```diff
@@ -0,0 +1,590 @@
...
+        on.
+
+        Args:
+            medrecord (MedRecord): The MedRecord.
+            attribute_name (MedRecordAttribute): The attribute name.
+
+        Returns:
+            MedRecordAttribute: Unique attribute name for the MedRecord.
+        """
+        created_name, counter = attribute_name, 1
```
JabobKrauskopf, line 119: should not be used except when functions return multiple values

```suggestion
        created_name = attribute_name,
        counter = 1
```

medmodels/data_synthesis/mtgan/modules/preprocessor.py
```diff
@@ -0,0 +1,590 @@
...
+                    (time - first_admission).days // time_interval_days
+                )
+                medrecord.edge[single_edge, concept_index_attribute] = (
+                    medrecord.select_nodes(
+                        node()
+                        .index()
+                        .is_in(list(medrecord.edge_endpoints(single_edge)))
+                        & node().in_group(self.concepts_group)
+                    )[0]
+                )
```
JabobKrauskopf, line 462: more spacing here

medmodels/data_synthesis/mtgan/modules/preprocessor.py
```diff
@@ -0,0 +1,590 @@
...
+            concept_index_attribute=concept_index_attribute,
+            number_of_windows_attribute=number_of_windows_attribute,
+            absolute_time_window_attribute=absolute_time_window_attribute,
+        )
+
+        return preprocessing_attributes
+
+    def _assign_concept_indices(
+        self, medrecord: MedRecord, concept_index_attribute: MedRecordAttribute
+    ) -> Dict[int, NodeIndex]:
```
JabobKrauskopf, line 491: Should probably be a List[NodeIndex] and the index of the concept is the index in the list. You could also create a type alias called `Concept` to make it clearer that it is a list of concepts

JabobKrauskopf (changes requested):
first rough review

medmodels/data_synthesis/mtgan/modules/preprocessor.py
MarIniOnz: Yes, realized there was no need for two attributes with the same functionality having different names.

medmodels/data_synthesis/mtgan/modules/preprocessor.py
MarIniOnz: hope it was enough.

medmodels/data_synthesis/mtgan/modules/preprocessor.py
MarIniOnz: okay! What happens is that the return of all that is a list! So the [0] returns always the only element of that list (there can only be one element, since we are referring to one edge).

medmodels/data_synthesis/mtgan/modules/preprocessor.py
MarIniOnz: Absolutely Hitchhikers guide to the Galaxy. I will make it optional though.

medmodels/data_synthesis/mtgan/modules/preprocessor.py
```diff
@@ -0,0 +1,593 @@
...
+            number_of_windows_attribute=number_of_windows_attribute,
+            absolute_time_window_attribute=absolute_time_window_attribute,
+        )
+
+        return preprocessing_attributes
+
+    def _assign_concept_indices(
+        self, medrecord: MedRecord, concept_index_attribute: MedRecordAttribute
+    ) -> ConceptsList:
+        """Assigns an index to each concept in the MedRecord and returns the index to concept dictionary.
```
LauraBoenchenLB, line 501: Not accurate anymore, returns a list of concepts

medmodels/data_synthesis/mtgan/modules/preprocessor.py
```diff
@@ -0,0 +1,593 @@
...
+        )
+        # Pruning the concepts after removing some edges and patient nodes
+        self._remove_uncommon_concepts(
+            medrecord,
+            minimum_number_ocurrences=1,
+        )
+        self._sample_patients(
+            medrecord, self.hyperparameters["number_of_sampled_patients"]
+        )
+        concepts_list = self._assign_concept_indices(medrecord, concept_index_attribute)
```
LauraBoenchenLB, line 585: Here we just have to be careful to not remove anything again if the concept_index is saved only based on the position in the list.

LauraBoenchenLB (changes requested):
Just some small requests and thoughts :)

medmodels/data_synthesis/mtgan/modules/preprocessor.py
MarIniOnz: Yes, that is important. Any idea, @JabobKrauskopf ? Or should we live with that?

JabobKrauskopf: is that a thing that can/should happen? I would have guessed that no concepts are removed once the concept mapping exists.

MarIniOnz: No concepts should be removed, nope.

JabobKrauskopf: then maybe add a comment that the index of the concept is based on its list index and hope you see it if anything should ever change ^^

MarIniOnz: I realized that having "concept_edge_attribute" erased was a mistake:
- The concept edge attribute holds the NAME of the concept ("Warfarin") in the EDGE connecting that concept to a patient node (easier querying afterwards but probably could be changed when the query engine appears)
- The concept index attribute holds the NUMERICAL INDEX of the concept (0, 1, 2, 3...)  in the concept NODE.

This sounds confusing to me too with the naming and so, so please feel free to suggest better naming.

medmodels/data_synthesis/mtgan/modules/preprocessor.py
```diff
@@ -0,0 +1,604 @@
...
+            directed=False,
+        )
+        time_windows = [
+            medrecord.edge[e, absolute_time_attribute] for e in patient_edges
+        ]
+        if min_codes_per_window == 0:
+            return time_windows
+
+        # Find which time windows appear less than min_codes_per_window times
+        time_window_counts = pd.Series(time_windows).value_counts()
```
LauraBoenchenLB, line 308: would using polars be faster for bigger data frames?

LauraBoenchenLB (dismissed):
Great job ⭐ Can't wait to see the full MTGAN in action!

JabobKrauskopf (changes requested):
Noice. So 🤏 close to being approved

JabobKrauskopf (approved):
image

Issue #160, opened 2024-08-09
Partial clone of MedRecord


MarIniOnz:
It would be useful to select here what should be copied by edges, nodes and/or groups.

LauraBoenchenLB:
It would be useful to select here what should be copied by edges, nodes and/or groups.

PR #204, opened 2024-09-17
feat: Implement simple clone

JabobKrauskopf:
Implements a simple `clone` method to basically deep copy a medrecord (not exactly that's why it's also not called deepcopy). Naming in accordance with rust and e.g. [polars](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.clone.html).

This also does not fulfill the issue requirement:
> It would be useful to select here what should be copied by edges, nodes and/or groups.

_Originally posted by @LauraBoenchenLB in https://github.com/limebit/medmodels/issues/160#issuecomment-2290946819_

Therefore this PR should not close issue #160


medmodels/medrecord/medrecord.py
```diff
@@ -1234,6 +1234,17 @@ def __getitem__(
 
         return self.select_edges(key)
 
+    def clone(self) -> "MedRecord":
```
MarIniOnz, line 1237: Why "MedRecord" and not MedRecord? Circular import workaround? In all of our modules we give as return the name of the class without the quotation marks so it would be nice to have a standard for it!

MarIniOnz (changes requested):
Simple, effective, and needed. What else do you want in a PR? 

Small comment but instant approve after that is solved.

medmodels/medrecord/medrecord.py
JabobKrauskopf: Idk why :D
Removed

MarIniOnz (approved):
_9e0e687b-4ed2-49cf-a3a4-71453aa0065d

LauraBoenchenLB (approved):
Yaaay, this will be so useful!! 😍

PR #209, opened 2024-09-17
fix: Renaming absolute risk function for correct one (absolute risk reduction)

MarIniOnz:
The previously named "absolute risk" function in the TreatmentEffect Estimate module was wrongly named like that. The implementation corresponded actually to "absolute risk reduction".

### DIFF
--- a/medmodels/treatment_effect/estimate.py
+++ b/medmodels/treatment_effect/estimate.py
@@ -318,19 +318,19 @@ class Estimate:
 
         return numerator / denominator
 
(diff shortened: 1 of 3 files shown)


LauraBoenchenLB (dismissed):
Great catch!

philippgrosser (approved):
Very nice! And good that you have also transferred the changes to NNT

JabobKrauskopf (approved):
noice :)

PR #211, opened 2024-09-17
feat: Adding contingency table print and treatment groups variables renaming

MarIniOnz:
Added a function that better visualizes how the different control and treated groups are separated. It is in `subject_counts`.

Also renamed the following variables everywhere for clarity purposes:
- treatment_true: "treatment_outcome_true"
- treatment_false: "treatment_outcome_false"
- control_true: "treatment_control_true"
- control_false: "treatment_control_false"

``` python
TretmentEffect.estimate.print_contingency_table(medrecord)
```
```
-----------------------------------
                   Outcome   
Group           True     False   
-----------------------------------
Treated         2        1       
Control         3        3       
-----------------------------------
```

### DIFF
--- a/medmodels/treatment_effect/continuous_estimators.py
+++ b/medmodels/treatment_effect/continuous_estimators.py
@@ -10,8 +10,8 @@ from medmodels.treatment_effect.temporal_analysis import find_reference_edge
 
 def average_treatment_effect(
     medrecord: MedRecord,
-    treatment_true_set: Set[NodeIndex],
-    control_true_set: Set[NodeIndex],
+    treatment_outcome_true_set: Set[NodeIndex],
+    control_outcome_true_set: Set[NodeIndex],
     outcome_group: Group,
     outcome_variable: MedRecordAttribute,
     reference: Literal["first", "last"] = "last",
@@ -39,10 +39,10 @@ def average_treatment_effect(
     Args:
         medrecord (MedRecord): An instance of the MedRecord class containing medical
             data.
-        treatment_true_set (Set[NodeIndex]): A set of node indices representing the
-            treated group.
-        control_true_set (Set[NodeIndex]): A set of node indices representing the
-            control group.
+        treatment_outcome_true_set (Set[NodeIndex]): A set of node indices representing
+            the treated group that also have the outcome.
+        control_outcome_true_set (Set[NodeIndex]): A set of node indices representing
+            the control group that have the outcome.
         outcome_group (Group): The group of nodes that contain the outcome variable.
         outcome_variable (MedRecordAttribute): The attribute in the edge that contains
             the outcome variable. It must be numeric and continuous.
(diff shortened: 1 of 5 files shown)


JabobKrauskopf (changes requested):
Why not return an actual object (custom object, not typed dictionary) at `subjects_contingency_table` and implement the `__str__` method on that object to pretty print it? That way the object is usable in basically the same way as the TypedDict and can automatically be pretty printed in the console or jupyter notebooks

Using `print` in a package is generally not a good idea because of the way stdio works differently on some machines (or is inaccessible). That's why you also generally don't see any `print_...` methods in other packages

medmodels/treatment_effect/continuous_estimators.py
```diff
@@ -39,9 +39,9 @@ def average_treatment_effect(
     Args:
         medrecord (MedRecord): An instance of the MedRecord class containing medical
             data.
-        treatment_true_set (Set[NodeIndex]): A set of node indices representing the
+        treatment_outcome_true_set (Set[NodeIndex]): A set of node indices representing the
             treated group.
```
philippgrosser, line 43: Isnt that the set that represents nodes that are true for outcome and treatment? Therefore the sentence _"A set of node indices representing the treated group."_ would be incorrect?

medmodels/treatment_effect/estimate.py
```diff
@@ -490,11 +484,72 @@ def cohens_d(
...
+            control_outcome_true_set=subjects.control_outcome_true,
             outcome_group=self._treatment_effect._outcomes_group,
             outcome_variable=outcome_variable,
             reference=reference,
             time_attribute=self._treatment_effect._time_attribute,
             add_correction=add_correction,
         )
+
+
+class TreatmentEffectGroups:
```
JabobKrauskopf, line 497: should probably be called contingency table

JabobKrauskopf (changes requested):
much better.

IMO we should keep the `subject_counts` method. The object that is returned from there should implement the `__str__` method that is implemented for `TreatmentEffectGroups` currently.
Then the `subjects_contingency_table` either returns the dictionary as before or keeps returning a custom object with another `__str__` function that does not print the number of subjects but the actual indices list.

JabobKrauskopf (approved):
PR APPROVED! But don’t get cocky, recruit! This is just one battle in the war of clean code. I expect nothing less than perfection next time. Now, get back out there and refactor like your IDE is on fire! OORAH /s

MarIniOnz:
> PR APPROVED! But don’t get cocky, recruit! This is just one battle in the war of clean code. I expect nothing less than perfection next time. Now, get back out there and refactor like your IDE is on fire! /s

AYE CAPTAIN!

philippgrosser (approved):
Very nice! Looking forward to use this!

Issue #165, opened 2024-08-14
Query Engine Refactor


JabobKrauskopf:
Completely remove the old query engine and write a new one from scratch with added funcitonality.

- [x] #170
- [x] #171
- [x] #172
- [x] #228 

___

After the above issues the first draft of the query engine will be merged.

- [x] Fix bugs
  - [x] #233 
- [x] Usability improvemets (sorted by importance & difficulty)
  - [x] #244
  - [x] #242
- [x] #289

___

`code example 1`
```python
a = edge.attribute("amount").min()
b = a.add(5)
```
After this, `a` should represent e.g. `5` and `b` should then represent 10

PR #226, opened 2024-10-09
refactor: update interface to reflect final (for now) query engine

JabobKrauskopf:
Updates in PR:
- fixed a few errors in interface
- interface now reflects actual query engine operations

Expected tests to fail (same as in the previous PR):
- `medmodels/medrecord/tests/test_builder.py`
- `medmodels/medrecord/tests/test_datatype.py`
- `medmodels/medrecord/tests/test_indexers.py`
- `medmodels/medrecord/tests/test_medrecord.py`
- `medmodels/medrecord/tests/test_overview.py`
- `medmodels/medrecord/tests/test_schema.py`
- `medmodels/treatment_effect/matching/tests/test_classic_distance_models.py`
- `medmodels/treatment_effect/matching/tests/test_metrics.py`
- `medmodels/treatment_effect/matching/tests/test_propensity_score.py`
- `medmodels/treatment_effect/tests/test_continuous_estimators.py`
- `medmodels/treatment_effect/tests/test_temporal_analysis.py`
- `medmodels/treatment_effect/tests/test_treatment_effect.py`


medmodels/medrecord/querying.pyi
```diff
@@ -86,24 +86,23 @@ class EdgeOperand:
...
-        self, query: Callable[[SingleValueOperand], None]
-    ) -> SingleValueOperand: ...
     def mean(self) -> SingleValueOperand: ...
     def median(self) -> SingleValueOperand: ...
     def mode(self) -> SingleValueOperand: ...
     def std(self) -> SingleValueOperand: ...
     def var(self) -> SingleValueOperand: ...
     def count(self) -> SingleValueOperand: ...
     def sum(self) -> SingleValueOperand: ...
+    def first(self) -> SingleValueOperand: ...
```
MarIniOnz, line 96: Why did we need a query argument in the previous first method and not anymore?

medmodels/medrecord/querying.pyi
```diff
@@ -214,6 +261,7 @@ class MultipleAttributesOperand:
     def trim_end(self) -> None: ...
     def lowercase(self) -> None: ...
     def uppercase(self) -> None: ...
+    def to_values(self) -> MultipleValuesOperand: ...
```
MarIniOnz, line 264: Does it make sense this is added only here?

medmodels/medrecord/querying.pyi
```diff
@@ -175,19 +179,62 @@ class SingleValueOperand:
     ) -> None: ...
     def clone(self) -> SingleValueOperand: ...
 
+class AttributesTreeOperand:
+    def max(self) -> MultipleAttributesOperand: ...
+    def min(self) -> MultipleAttributesOperand: ...
+    def count(self) -> MultipleAttributesOperand: ...
```
MarIniOnz, line 185: no mean, mode, median, std or var here? I can easily think of the reason but just wanted you to check it just in case

medmodels/medrecord/querying.pyi
```diff
@@ -175,19 +179,62 @@ class SingleValueOperand:
...
+    def sum(self) -> MultipleAttributesOperand: ...
+    def first(self) -> MultipleAttributesOperand: ...
+    def last(self) -> MultipleAttributesOperand: ...
+    def is_string(self) -> None: ...
+    def is_int(self) -> None: ...
```
MarIniOnz, line 190: why no is_float, is_datetime, is_null or is_bool here?

medmodels/medrecord/querying.pyi
```diff
@@ -175,19 +179,62 @@ class SingleValueOperand:
...
+    def is_not_in(self, values: MultipleAttributesComparisonOperand) -> None: ...
+    def starts_with(self, value: SingleAttributeComparisonOperand) -> None: ...
+    def ends_with(self, value: SingleAttributeComparisonOperand) -> None: ...
+    def contains(self, value: SingleAttributeComparisonOperand) -> None: ...
+    def add(self, value: SingleAttributeComparisonOperand) -> None: ...
+    def subtract(self, value: SingleAttributeComparisonOperand) -> None: ...
+    def multiply(self, value: SingleAttributeComparisonOperand) -> None: ...
+    def modulo(self, value: SingleAttributeComparisonOperand) -> None: ...
+    def power(self, value: SingleAttributeComparisonOperand) -> None: ...
+    def absolute(self) -> None: ...
```
MarIniOnz, line 213: no sqrt, round, ceil or floor?

MarIniOnz (commented):
Awesooooooome clooooooooooose

medmodels/medrecord/querying.pyi
JabobKrauskopf: Because it was unnecessarily difficult to implement and this should have the same functionality, e.g.
```python
values = node.attribute("time")

value = values.first_where(lambda value: value.is_datetime())
```
should be the same thing as
```python
values = node.attribute("time")

values.is_datetime()

value = values.first()
```

medmodels/medrecord/querying.pyi
JabobKrauskopf: A `MedRecordAttribute` can only be a `string` or `int`

medmodels/medrecord/querying.pyi
JabobKrauskopf: IMO it just doesn't make sense to get the `mean`, `mode`, `median`, `std` or `var` of a node (or edge) attribute. Most of them will be strings anyway which wouldn't even support these operations

medmodels/medrecord/querying.pyi
JabobKrauskopf: disregard. it should also not have a `round`, `ceil` and `floor`. A `MedRecordAttribute` is either a `string` or a `int`. On strings these operations would not do anything and on an `int` as well since an integer is always rounded

medmodels/medrecord/querying.pyi
MarIniOnz: open small issue/add to a previous one so its not forgotten

JabobKrauskopf: added to #165

MarIniOnz (approved):

PR #230, opened 2024-10-11
refactor: stricter linting rules

FloLimebit:
This branch introduces the following changes: 

* Ruff linting rules in pyproject.toml file made explicit and as strict as possible/reasonable
  * each linting rules has a short code comment explaining what the rule is about 
  * an exclusion list was added to avoid conflicts with the auto formatter
  * the ruff option `preview = true` got activated -> ruff supports supports now potentially unsafe fixes.
    * since we have an extremely high test coverage, I'd suggest we go with this option and see if it breaks something down the road 
* fix for generating API docs of treatment effect module
* small improvement to `docs/conf.py` ensuring that `__all__` definitions are respected
* Docstrings Developer Guide got extended to be more explicit and explanatory


medmodels/medrecord/datatype.py
```diff
@@ -212,8 +211,9 @@ class Union(DataType):
 
     def __init__(self, *dtypes: DataType) -> None:
         if len(dtypes) < 2:
-            raise ValueError("Union must have at least two arguments")
-        elif len(dtypes) == 2:
+            msg = "Union must have at least two arguments"
+            raise ValueError(msg)
```
JabobKrauskopf, line 215: do you know which linting rule this is? Not sure if I like this

FloLimebit: When adding the string into the raise expression directly, it will be shown twice on error

https://docs.astral.sh/ruff/rules/raw-string-in-exception/

FloLimebit: But let's collect and discuss linting rules we might not wanna subscribe to

JabobKrauskopf: with the explanation this rule makes sense

JabobKrauskopf (dismissed):
image

docs/developer_guide/docstrings.md
```diff
@@ -221,26 +214,28 @@ Document parameters under the `Args` section. Each parameter should include its
...
+
+When writing type definitions in argument docstrings, avoid placing line breaks inside the type annotations. For instance, complex types like `(Union[str, List[str], Dict[str, Any]])` should appear on a single line without splitting across lines. This ensures the type definition remains clear and avoids parsing issues.
+
+If a type definition exceeds the maximum line length limit, deactivate the line length rule for that doctstring using [`# noqa: E501`](https://docs.astral.sh/ruff/rules/line-too-long/#error-suppression) after the closing `"""`. This keeps the type annotation intact while preventing the linter from flagging the long line as an error.
 :::
 
+
+:::{admonition} No Boolean Argument Types
+:type: note
+Avoid using booleans as function arguments due to the "boolean trap." Booleans reduce code clarity, as `True` or `False` doesn't explain meaning. They also limit future flexibility — if more than two options are needed, changing the argument type can cause breaking changes. Instead, use an enum or a more descriptive type for better clarity and flexibility.
```
MarIniOnz, line 233:
```suggestion
Avoid using booleans as function arguments due to the "boolean trap". Booleans reduce code clarity, as `True` or `False` doesn't explain meaning. They also limit future flexibility — if more than two options are needed, changing the argument type can cause breaking changes. Instead, use an enum or a more descriptive type for better clarity and flexibility.
```

docs/developer_guide/example_docstrings.py
```diff
@@ -0,0 +1,58 @@
+from __future__ import annotations
+
+from typing import Any, Dict, Iterator
+
+
+def example_function_args(
+    param1: int,
+    param2: str | int,
+    optional_param: list[str] | None = None,
```
MarIniOnz, line 9: We are using the List typing from "typing" in the whole repository, instead of the native list from python. Maybe it would be nice to share that across the codebase.

docs/developer_guide/example_docstrings.py
```diff
@@ -0,0 +1,58 @@
...
+    *args: float | str,
+    **kwargs: Dict[str, Any],
+) -> tuple[bool, list[str]]:
```
MarIniOnz, line 12: same as above, do not forget also to change in docstrings.

medmodels/treatment_effect/matching/tests/test_classic_distance_models.py
```diff
@@ -2,22 +2,23 @@
...
 
 from medmodels.treatment_effect.matching.algorithms import (
     classic_distance_models as cdm,
 )
 
 
 class TestClassicDistanceModels(unittest.TestCase):
-    def test_nearest_neighbor(self):
+    def test_nearest_neighbor(self) -> None:
         ###########################################
```
MarIniOnz, line 14: Missing function docstrings (not your fault but we could include it here hehe)

medmodels/medrecord/medrecord.py
```diff
@@ -778,26 +776,33 @@ def add_edges(
...
-            self.add_edges_to_group(group, edge_indices)
+        self.add_edges_to_group(group, edge_indices)
 
+        if group is None:
             return edge_indices
 
+        if not self.contains_group(group):
+            self.add_group(group)
+
+        self.add_edges_to_group(group, edge_indices)
```
MarIniOnz, line 802: This was added twice.

MarIniOnz (changes requested):
Awesome PR!
Some small corrections.

Also, when performing `make lint` and `make format` this error pops up. Is it expected? We should probably fix it before merging:
```ruff failed
  Cause: Failed to parse /Users/martin.iniguez/Desktop/projects/medmodels/pyproject.toml
  Cause: TOML parse error at line 90, column 1
   |
90 | [tool.ruff.lint]
   | ^^^^^^^^^^^^^^^^
Unknown rule selector: `DOC`
```

JabobKrauskopf:
@MarIniOnz 

> Also, when performing make lint and make format this error pops up. Is it expected? We should probably fix it before merging:

I don't get this error. Could you try running `make clean` and then e.g. `make format` again?

MarIniOnz (approved):
DALLE2024-10-1514 14 23-Aroughsketchydrawingofasealgivingathumbsup Surroundingthesealarevariouscodingbugsanderrorsrepresentedaschaoticsymbolslikeexc-ezgif com-webp-to-jpg-converter

medmodels/treatment_effect/matching/propensity.py
```diff
@@ -78,6 +80,10 @@ def match_controls(
             Set[NodeIndex]:  Node Ids of the matched controls.
         """
         # Preprocess the data
+        if one_hot_covariates is None:
```
LauraBoenchenLB, line 83: In general, how will this work in the documentation? I get that lists are bad as an input because they are mutable and it could cause problems when calling the function multiple times, but will we just have the default values mentioned in the docstring like it is here?

medmodels/treatment_effect/treatment_effect.py
```diff
@@ -11,24 +11,26 @@
...
-    NodeIndex,
-)
 from medmodels.treatment_effect.builder import TreatmentEffectBuilder
 from medmodels.treatment_effect.estimate import Estimate
-from medmodels.treatment_effect.matching.algorithms.propensity_score import Model
-from medmodels.treatment_effect.matching.matching import MatchingMethod
 from medmodels.treatment_effect.report import Report
 from medmodels.treatment_effect.temporal_analysis import find_node_in_time_window
 
+if TYPE_CHECKING:
```
LauraBoenchenLB, line 22: All imports that are used for typing have to be prefaced with this?

LauraBoenchenLB (dismissed):
Great job! Interested to see how this will affect our coding in the future. Just a loooot of questions from my side, they don't hinder approval.

docs/developer_guide/example_docstrings.py
```diff
@@ -0,0 +1,58 @@
+from __future__ import annotations
+
+from typing import Any, Dict, Iterator, List
+
+
+def example_function_args(
+    param1: int,
+    param2: str | int,
+    optional_param: List[str] | None = None,
```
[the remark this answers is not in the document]
JabobKrauskopf: this file had mostly default python typing. `Optional`, `Tuple` etc. is just syntactic sugar for the default typing but we like sugar. I adjusted the example to use the `typing` stuff

medmodels/medrecord/schema.py
```diff
@@ -31,48 +32,45 @@ def _from_pyattributetype(py_attribute_type: PyAttributeType) -> AttributeType:
...
-        Converts an AttributeType to a PyAttributeType.
+        """Converts an AttributeType to a PyAttributeType.
 
         Returns:
             PyAttributeType: The converted PyAttributeType.
         """
         if self == AttributeType.Categorical:
             return PyAttributeType.Categorical
-        elif self == AttributeType.Continuous:
+        if self == AttributeType.Continuous:
```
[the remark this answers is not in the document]
JabobKrauskopf: this is probably the "recursive" version of https://docs.astral.sh/ruff/rules/superfluous-else-return/

this is not super readable now but it should be a `match` statement anyway. However this is only supported from python 3.10 upwards and we support 3.9 which me might not need to support imo

medmodels/treatment_effect/matching/propensity.py
JabobKrauskopf: imo this does not need to be specially documented. It does not matter to the user wether the default value is a `[]` or a `None`. The user just needs to know that he/she can pass a list or nothing and that is achieved in both versions of the typing

MarIniOnz (approved):
DALLE2024-10-1514 14 02-Averysketchyhand-drawnstyleillustrationofasealgivingathumbsup Aroundthesealtherearevariouscodingbugsanderrormessageslikesy-ezgif com-webp-to-jpg-converter

docs/developer_guide/example_docstrings.py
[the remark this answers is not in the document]
LauraBoenchenLB: sure 🍬

medmodels/treatment_effect/matching/propensity.py
LauraBoenchenLB: In case of an empty list, yes. But here for example we have lists full of default values that will be set in case of no input.

LauraBoenchenLB: Maybe we should remove these default values completely anyways, I'm not sure if it makes sense, since Attribute naming is so free. But that is not the topic of this PR.

JabobKrauskopf: You're right, sorry. Was on my phone and didn't check the whole code.
Then I'll pass this on for @FloLimebit to answer once he's back from his vacation.

PR #233, opened 2024-10-14
fix: comparison operations now correctly evaluate operand

JabobKrauskopf:
This fixes a bug where operations that change the value of an operand (e.g. `.add(...)`, `.subtract(...)`, `.round()` etc.) are not applied correctly when the operand is used in a comparison


MarIniOnz (approved):
So fast. Well done Jakob!
DALLE2024-10-1510 33 09-Asealdressedasaprofessorwearingglassesandasuitgivingathumbsup Behindthesealisalargeblackboardfilledwithcomplexmathematical-ezgif com-webp-to-jpg-converter

PR #239, opened 2024-10-16
fix: linting issues in medrecord module

JabobKrauskopf:
This PR should fix the linting issues in the medrecord module.

I used `# noqa: C901` a few times for function definitions that just are very complex.

I also used `# noqa: ANN201` in the `schema.py` on

- `def keys(self):` because I couldn't get the actual return type `dict_keys`
- `def values(self):` because I couldn't get the actual return type `dict_values`
- `def items(self):` because I couldn't get the actual return type `dict_items`

I didn't adjust any docstrings. There is an issue for that #240


MarIniOnz (changes requested):
All good but in medrecord/medrecord.py file, there are still the following linting errors:
boolean-typed positional argument in function definition (for lines 1200-1213)

MarIniOnz (approved):
DALLE2024-10-1713 59 24-AdetaileddepictionofaNavySEALinfulltacticalgeargivingaformalmilitarysalute TheSEALwearsacamouflageuniformwithahelmetgoggles-ezgif com-webp-to-jpg-converter

PR #243, opened 2024-10-17
feat: implement exclude operation

JabobKrauskopf:
I implemented a exclude operation for basically all operands.

Example:

```python
def query(node: NodeOperand) -> None:
    node.exclude(lambda node: node.attribute("age").greater_than(30))
```

I also fixed a tiny issue with `either_or` in `NodeOperation` and `EdgeOperation`


MarIniOnz (approved):
DALLE2024-10-1716 06 26-ANavySEALsoldierinuniformhuggingananimatedseal Theanimatedsealshouldhaveafriendlyandcartoonishappearancewithlargeeyesandasmi-ezgif com-webp-to-jpg-converter

PR #246, opened 2024-10-17
feat: implement duration

JabobKrauskopf:
I implemented support for duration (a.k.a. time delta).
Tests are missing and will be added when I'm back from vacation


MarIniOnz (approved):
Amazing PR so fast!

#247 will fix my only concern about it: No negative timedeltas can be used so far

DALLE2024-10-1814 25 02-AgroupofsealsthemarineanimalsdressedasNavySEALswearingnavyblueuniformstacticalvestsandmilitarycapswithbadges Theyarestand-ezgif com-webp-to-jpg-converter

PR #248, opened 2024-10-18
refactor: optimize treatment effect and associated functions with new query engine

MarIniOnz:
- Eliminated `find_node_in_time_window` (not needed anymore in that format) and refactored into the `TreatmentEffect` class as "_query_node_within_time_window".
- Changed "treated_group" and "control_group" variables into "treated_set" and "control_set" in all TEE files since they might lead to confusion. Also, changed a couple variable names that should have been changed in previous PRs but were forgotten.
- Changed datatypes and error messages in tests for new query.

Note:
    - Some errors used as sanity checks were eliminated from the TreatmentEffect because of the implementation of the Query Engine. These should be re-implemented after issue #100

### DIFF
--- a/medmodels/treatment_effect/estimate.py
+++ b/medmodels/treatment_effect/estimate.py
@@ -164,10 +164,13 @@ class Estimate:
                 MedRecord (patients, treatments, outcomes).
         """
         self._check_medrecord(medrecord=medrecord)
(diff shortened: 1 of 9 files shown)


JabobKrauskopf (approved):
image

LauraBoenchenLB (approved):
GREEEEEAAAAT 🦭

PR #256, opened 2024-11-19
fix: the way essential covariates and one-hot-encoded ones are processed

MarIniOnz:
- Essential covariates are no longer "gender" and "age" by default, but all attributes from the patients group. Raises errors in case there are not enough control/treated patients with those attributes.
- One-hot-encoded variables are not "gender" by default, but all attributes characterised by `extract_attribute_summary` function as Categorical.
- Added tests to the Matching Class

- Changed some variable names (hyperparam -> hyperparameters) for better variable naming consistency.

### DIFF
--- a/medmodels/treatment_effect/matching/matching.py
+++ b/medmodels/treatment_effect/matching/matching.py
@@ -55,9 +55,11 @@ class Matching(ABC):
             treated_set (Set[NodeIndex]): Set of control subjects.
             patients_group (Group): The group of patients.
             essential_covariates (Optional[MedRecordAttributeInputList]):
-                Covariates that are essential for matching. Defaults to None.
+                Covariates that are essential for matching. Defaults to None, meaning
+                all the attributes of the patients are used.
             one_hot_covariates (Optional[MedRecordAttributeInputList]):
-                Covariates that are one-hot encoded for matching. Defaults to None.
+                Covariates that are one-hot encoded for matching. Defaults to None,
+                meaning all the categorical attributes of the patients are used.
 
         Returns:
             Tuple[pl.DataFrame, pl.DataFrame]: Treated and control groups with their
@@ -74,8 +76,6 @@ class Matching(ABC):
                     medrecord.node[medrecord.nodes_in_group(patients_group)]
                 )
             )
-        else:
-            essential_covariates = list(essential_covariates)
 
         control_set = self._check_nodes(
             medrecord=medrecord,
@@ -85,9 +85,9 @@ class Matching(ABC):
         )
 
         if "id" not in essential_covariates:
-            essential_covariates.append("id")
+            essential_covariates.append("id")  # pyright: ignore[reportArgumentType]
 
-        # Dataframe wth the essential covariates
+        # Dataframe with the essential covariates
         data = pl.DataFrame(
             data=[
                 {"id": k, **v}
@@ -104,11 +104,21 @@ class Matching(ABC):
             one_hot_covariates = [
                 covariate
                 for covariate, values in attributes.items()
-                if "values" in values
+                if "Categorical" in values["type"]
             ]
 
-        if not all(
-            covariate in essential_covariates for covariate in one_hot_covariates
+            one_hot_covariates = [
+                covariate
+                for covariate in one_hot_covariates
+                if covariate in essential_covariates
+            ]
+
+        # If there are one-hot covariates, check if all are in the essential covariates
+        if (
+            not all(
+                covariate in essential_covariates for covariate in one_hot_covariates
+            )
+            and one_hot_covariates
         ):
             msg = "One-hot covariates must be in the essential covariates"
             raise AssertionError(msg)
@@ -122,8 +132,8 @@ class Matching(ABC):
 
         # Add to essential covariates the new columns created by one-hot encoding and
         # delete the ones that were one-hot encoded
-        essential_covariates.extend(new_columns)
-        [essential_covariates.remove(col) for col in one_hot_covariates]
+        essential_covariates.extend(new_columns)  # pyright: ignore[reportArgumentType]
+        [essential_covariates.remove(col) for col in one_hot_covariates]  # pyright: ignore[reportArgumentType]
         data = data.select(essential_covariates)
 
         # Select the sets of treated and control subjects
@@ -170,8 +180,15 @@ class Matching(ABC):
                 lambda node: query_essential_covariates(node, control_set)
             )
         )
+
         if len(control_set) < self.number_of_neighbors * len(treated_set):
-            msg = "Not enough control subjects to match the treated subjects"
+            msg = (
+                f"Not enough control subjects to match the treated subjects. "
+                f"Number of controls: {len(control_set)}, "
+                f"Number of treated subjects: {len(treated_set)}, "
+                f"Number of neighbors required per treated subject: {self.number_of_neighbors}, "
+                f"Total controls needed: {self.number_of_neighbors * len(treated_set)}."
+            )
             raise ValueError(msg)
 
         if len(treated_set) != len(
--- a/medmodels/treatment_effect/matching/matching.py
+++ b/medmodels/treatment_effect/matching/matching.py
@@ -8,7 +8,16 @@ score matching and nearest neighbor matching.
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, Literal, Optional, Set, Tuple, TypeAlias
+from typing import (
+    TYPE_CHECKING,
+    List,
+    Literal,
+    Optional,
+    Sequence,
+    Set,
+    Tuple,
+    TypeAlias,
+)
 
 import polars as pl
 
@@ -18,7 +27,7 @@ from medmodels.medrecord.medrecord import MedRecord
 if TYPE_CHECKING:
     from medmodels.medrecord.medrecord import MedRecord
     from medmodels.medrecord.querying import NodeOperand
-    from medmodels.medrecord.types import Group, MedRecordAttributeInputList, NodeIndex
+    from medmodels.medrecord.types import Group, MedRecordAttribute, NodeIndex
 
 MatchingMethod: TypeAlias = Literal["propensity", "nearest_neighbors"]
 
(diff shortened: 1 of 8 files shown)


medmodels/treatment_effect/matching/matching.py
```diff
@@ -79,13 +131,66 @@ def _preprocess_data(
...
+            treated_set (Set[NodeIndex]): Set of treated subjects.
+            control_set (Set[NodeIndex]): Set of control subjects.
+            essential_covariates (MedRecordAttributeInputList): Covariates that are
+                essential for matching.
+
+        Returns:
+            Set[NodeIndex]: The control set.
+
+        Raises:
+            ValueError: If not enough control subjects to match the treated subjects.
```
LauraBoenchenLB, line 154: also raises `ValueError: If some treated nodes do not have all the essential covariates.`

medmodels/treatment_effect/matching/matching.py
```diff
@@ -79,13 +131,66 @@ def _preprocess_data(
...
+            node.index().is_in(list(patients_set))
+
+        control_set = set(
+            medrecord.select_nodes(
+                lambda node: query_essential_covariates(node, control_set)
+            )
+        )
+        if len(control_set) < self.number_of_neighbors * len(treated_set):
+            raise ValueError(
+                "Not enough control subjects to match the treated subjects"
```
LauraBoenchenLB, line 173: Maybe add the number of neighbors in the error message?

medmodels/treatment_effect/matching/matching.py
```diff
@@ -79,13 +131,66 @@ def _preprocess_data(
...
+            )
+
+        if len(treated_set) != len(
```
LauraBoenchenLB, line 176: I would maybe add this query first, so the control node query can be stopped earlier if not all treated  nodes have the essential covariates.

medmodels/treatment_effect/matching/tests/test_matching.py
```diff
@@ -0,0 +1,241 @@
...
+        )
+
+        # Assert that the matched controls are a subset of the control set
+        self.assertTrue(matched_controls.issubset(control_set))
+
+        # Assert that the correct number of controls were matched
+        self.assertEqual(len(matched_controls), len(treated_set))
+
+        # Assert it works equally if no covariates are given (automatically assigned)
+        matched_controls_no_covariates_specified = neighbors_matching.match_controls(
```
LauraBoenchenLB, line 148: Maybe add checking if `matched_controls_no_covariates_specified` and `matched_controls ` are equal?

medmodels/treatment_effect/matching/tests/test_matching.py
```diff
@@ -0,0 +1,241 @@
...
+        with self.assertRaises(ValueError) as context:
+            neighbors_matching._check_nodes(
+                medrecord=self.medrecord,
+                treated_set={"P2", "P10"},
+                control_set=control_set,
+                essential_covariates=["age", "gender"],
+            )
+        self.assertEqual(
+            str(context.exception),
+            "Some treated nodes do not have all the essential covariates",
```
LauraBoenchenLB, line 217: Is this correct? Shouldn't it be an Error because 'P10' doesn't exist instead of it not having the covariate

LauraBoenchenLB (changes requested):
Looks great! A few changes needed, but overall very well executed!

medmodels/treatment_effect/matching/tests/test_matching.py
MarIniOnz: Changed to include the P10 inside the `MedRecord` without covariates

medmodels/treatment_effect/matching/matching.py
```diff
@@ -60,6 +94,24 @@ def _preprocess_data(
         )
         original_columns = data.columns
 
+        if one_hot_covariates is None:
```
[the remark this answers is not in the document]
MarIniOnz: Very good point: I will add then the considerations for this to not happen. Thanks for the very insightful remark

LauraBoenchenLB (dismissed):
Looks great! All my concerns have been addressed, very happy with the result! 🦭

LauraBoenchenLB (approved):
Let's meeeerge 🦭

medmodels/treatment_effect/matching/matching.py
```diff
@@ -29,36 +33,68 @@
...
+
+        Raises:
+            AssertionError: If the one-hot covariates are not in the essential
+                covariates.
         """
-        essential_covariates = [str(covariate) for covariate in essential_covariates]
+        if essential_covariates is None:
+            # If no essential covariates provided, use all attributes of patients group
+            essential_covariates = list(
+                extract_attribute_summary(
```
JabobKrauskopf, line 82: this method is way too expensive for this use-case. rather use something like `{key for subdict in attribute_dictionary.values() for key in subdict.keys()}` This would also return a set of all possible attributes

medmodels/treatment_effect/matching/matching.py
```diff
@@ -86,6 +149,65 @@ def _preprocess_data(
...
+            ValueError: If not enough control subjects to match the treated subjects.
+            ValueError: If some treated nodes do not have all the essential covariates.
+        """
+
+        def query_essential_covariates(
+            node: NodeOperand, patients_set: Set[NodeIndex]
+        ) -> None:
+            """Query the nodes that have all the essential covariates."""
+            for attribute in essential_covariates:
+                node.has_attribute(attribute)
```
JabobKrauskopf, line 181: if you use the `List[MedRecordAttribute]` typing in line 157, this can be rewritten as
```suggestion
            node.has_attribute(essential_covariates)
```

JabobKrauskopf (changes requested):

medmodels/treatment_effect/matching/neighbors.py
```diff
@@ -36,27 +36,29 @@ def __init__(
...
-            essential_covariates (Optional[MedRecordAttributeInputList], optional):
+            treated_set (Set[NodeIndex]): Set of treated subjects.
+            patients_group (Group): Group of patients in the MedRecord.
+            essential_covariates (Optional[Sequence[MedRecordAttribute]], optional):
                 Covariates that are essential for matching. Defaults to
                 ["gender", "age"].
-            one_hot_covariates (Optional[MedRecordAttributeInputList], optional):
+            one_hot_covariates (Optional[Sequence[MedRecordAttribute]], optional):
                 Covariates that are one-hot encoded for matching. Defaults to
                 ["gender"].
```
LauraBoenchenLB, line 63: Defaults shouldn't be in the description anymore

medmodels/treatment_effect/matching/propensity.py
```diff
@@ -40,43 +41,44 @@ def __init__(
...
         Args:
             medrecord (MedRecord): medrecord object containing the data.
-            treated_set (Set[NodeIndex]): Set of treated subjects.
             control_set (Set[NodeIndex]): Set of control subjects.
-            essential_covariates (Optional[MedRecordAttributeInputList], optional):
+            treated_set (Set[NodeIndex]): Set of treated subjects.
+            patients_group (Group): Group of patients in MedRecord.
+            essential_covariates (Optional[Sequence[MedRecordAttribute]], optional):
                 Covariates that are essential for matching. Defaults to
                 ["gender", "age"].
```
LauraBoenchenLB, line 80: see above

LauraBoenchenLB (changes requested):
Minor detail change request!

LauraBoenchenLB (approved):
Perfect! This get's an approval 🤩

medmodels/treatment_effect/treatment_effect.py
```diff
@@ -68,8 +68,8 @@ class TreatmentEffect:
     _filter_controls_query: Optional[NodeQuery]
 
     _matching_method: Optional[MatchingMethod]
-    _matching_essential_covariates: MedRecordAttributeInputList
-    _matching_one_hot_covariates: MedRecordAttributeInputList
+    _matching_essential_covariates: Optional[MedRecordAttributeInputList]
+    _matching_one_hot_covariates: Optional[MedRecordAttributeInputList]
```
JabobKrauskopf, line 72: why is this optional? `_set_configuration` sets defaults for these values, so they can never be `None`

medmodels/treatment_effect/matching/neighbors.py
```diff
@@ -36,44 +36,44 @@ def __init__(
...
-                ["gender", "age"].
-            one_hot_covariates (Optional[MedRecordAttributeInputList], optional):
-                Covariates that are one-hot encoded for matching. Defaults to
-                ["gender"].
+            treated_set (Set[NodeIndex]): Set of treated subjects.
+            patients_group (Group): Group of patients in the MedRecord.
+            essential_covariates (Optional[List[MedRecordAttribute]], optional):
+                Covariates that are essential for matching. Defaults to None, meaning
+                all the attributes of the patients are used.
+            one_hot_covariates (Optional[List[MedRecordAttribute]], optional):
```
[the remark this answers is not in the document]
JabobKrauskopf: because of `Sequence` vs `List`

JabobKrauskopf (approved):
I told ChatGPT to surprise me. This is what it came up with. I'm scared.

Screen Shot 2025-01-21 at 11 45 19

MarIniOnz:
> I told ChatGPT to surprise me. This is what it came up with. I'm scared.
> 
> The seals won the Seal-Human war.
> 
> Screen Shot 2025-01-21 at 11 45 19

that is the future.

JabobKrauskopf (approved):
This is apparently the continuation of the story

Screen Shot 2025-01-21 at 11 58 34

LauraBoenchenLB (commented):
Tiny question, will approve after! <3

medmodels/treatment_effect/matching/matching.py
```diff
@@ -8,50 +8,94 @@
...
     def _preprocess_data(
         self,
         *,
         medrecord: MedRecord,
         control_set: Set[NodeIndex],
         treated_set: Set[NodeIndex],
-        essential_covariates: MedRecordAttributeInputList,
-        one_hot_covariates: MedRecordAttributeInputList,
+        patients_group: Group,
+        essential_covariates: Optional[List[MedRecordAttribute]] = None,
```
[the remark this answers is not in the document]
MarIniOnz: I would say it is okay, since it will always be a list. It is just a problem with typing in the other functions and that is why `Sequence` was recquired in the other one. Right, @JabobKrauskopf?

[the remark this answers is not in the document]
JabobKrauskopf: more or less, yes

PR #270, opened 2024-12-02
refactor: fast-computed representation of the MedRecord when in debugging mode

MarIniOnz:
The MedRecord can have loads of different nodes and groups, so going through their attributes can be very slow work. This can hinder the performance of the MedRecord in debugging mode. For that reason, a short representation of the MedRecord will be shown in debugging mode> 

`<MedRecord: 73 nodes, 160 edges>`

### DIFF
--- a/medmodels/medrecord/medrecord.py
+++ b/medmodels/medrecord/medrecord.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import sys
 from typing import Callable, Dict, List, Optional, Sequence, Union, overload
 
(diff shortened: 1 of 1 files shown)


philippgrosser (approved):

LauraBoenchenLB (approved):
Very good idea, super useful for speed ⚡

JabobKrauskopf (approved):
This feels a bit weird and hacky. Approved but I'll add an issue to check for proper debug support

PR #274, opened 2024-12-05
feat: making static analysis as the default option for the TEE

MarIniOnz:
By setting the time_attribute to None, there is no consideration for time dynamics now in the TEE, just whether there are any edges between the treatments, outcomes and the patient nodes.


medmodels/treatment_effect/continuous_estimators.py
```diff
@@ -50,46 +51,76 @@ def average_treatment_effect(
...
+            ]
+        )
+
+    else:
+        edges_treated_outcomes = medrecord.select_edges(
+            lambda edge: query_edges_set_outcome(
+                edge, treatment_outcome_true_set, outcome_group
+            )
+        )
+        treated_outcomes = treated_outcomes = np.array(
```
LauraBoenchenLB, line 102:
```suggestion
        treated_outcomes = np.array(
```

is this maybe a typo or what does that mean?

medmodels/treatment_effect/continuous_estimators.py
```diff
@@ -50,46 +51,76 @@ def average_treatment_effect(
...
+            lambda edge: query_edges_set_outcome(
+                edge, control_outcome_true_set, outcome_group
+            )
+        )
+        control_outcomes = treated_outcomes = np.array(
+            [
+                medrecord.edge[edge_id][outcome_variable]
+                for edge_id in edges_control_outcomes
+            ]
+        )
```
LauraBoenchenLB, line 119:
```suggestion
        control_outcomes = np.array(
            [
                medrecord.edge[edge_id][outcome_variable]
                for edge_id in edges_control_outcomes
            ]
        )
```

medmodels/treatment_effect/continuous_estimators.py
```diff
@@ -207,3 +266,34 @@ def cohens_d(
...
+    """
+    list_nodes = list(set)
+    edge.either_or(
+        lambda edge: edge.source_node().index().is_in(list_nodes),
+        lambda edge: edge.target_node().index().is_in(list_nodes),
+    )
+    edge.either_or(
+        lambda edge: edge.source_node().index().is_in(list_nodes),
+        lambda edge: edge.target_node().index().is_in(list_nodes),
+    )
```
LauraBoenchenLB, line 290: double?

medmodels/treatment_effect/continuous_estimators.py
```diff
@@ -207,3 +266,34 @@ def cohens_d(
...
+
+    edge.either_or(
+        lambda edge: edge.source_node().in_group(outcomes_group),
+        lambda edge: edge.target_node().in_group(outcomes_group),
+    )
+    edge.either_or(
+        lambda edge: edge.source_node().in_group(outcomes_group),
+        lambda edge: edge.target_node().in_group(outcomes_group),
+    )
```
LauraBoenchenLB, line 299: Also double? Or does this have a purpose?

medmodels/treatment_effect/continuous_estimators.py
```diff
@@ -141,8 +172,11 @@ def cohens_d(
             exposure time. Options include "first" and "last". If "first", the function
             returns the earliest exposure edge. If "last", the function returns the
             latest exposure edge. Defaults to "last".
-        time_attribute (MedRecordAttribute, optional): The attribute in the edge that
-            contains the time information. Defaults to "time".
+        time_attribute (Optional[MedRecordAttribute], optional): The attribute in the
+            edge that contains the time information. If it is equal to None, there is
+            no time component in the data and all edges between the sets and the
+            outcomes are considered for the average treatment effect. Defaults to
```
LauraBoenchenLB, line 178:
```suggestion
            outcomes are considered for the Cohen's D calculation. Defaults to
```

medmodels/treatment_effect/continuous_estimators.py
```diff
@@ -153,37 +187,62 @@ def cohens_d(
...
+                for node_index in control_outcome_true_set
+            ]
+        )
+    else:
+        edges_treated_outcomes = medrecord.select_edges(
+            lambda edge: query_edges_set_outcome(
+                edge, treatment_outcome_true_set, outcome_group
+            )
+        )
+        treated_outcomes = treated_outcomes = np.array(
```
LauraBoenchenLB, line 225: see above

medmodels/treatment_effect/continuous_estimators.py
```diff
@@ -153,37 +187,62 @@ def cohens_d(
...
+                for edge_id in edges_treated_outcomes
+            ]
+        )
+
+        edges_control_outcomes = medrecord.select_edges(
+            lambda edge: query_edges_set_outcome(
+                edge, control_outcome_true_set, outcome_group
+            )
+        )
+        control_outcomes = treated_outcomes = np.array(
```
LauraBoenchenLB, line 237: see above

medmodels/treatment_effect/continuous_estimators.py
```diff
@@ -153,37 +187,62 @@ def cohens_d(
...
+        treated_outcomes = treated_outcomes = np.array(
+            [
+                medrecord.edge[edge_id][outcome_variable]
+                for edge_id in edges_treated_outcomes
+            ]
+        )
+
+        edges_control_outcomes = medrecord.select_edges(
+            lambda edge: query_edges_set_outcome(
+                edge, control_outcome_true_set, outcome_group
```
LauraBoenchenLB, line 234: I'm not 100% sure that I understand the use of this. If the set is already treatment/control and outcome_true, why are we checking again if they are connected to the outcome group? Or is this just the naming of the variables that is confusing to me?

medmodels/treatment_effect/treatment_effect.py
```diff
@@ -110,10 +114,12 @@ def _set_configuration(
...
                 Defaults to "patients".
-            time_attribute (MedRecordAttribute, optional): The time attribute to use for
-                time-based analysis. Defaults to "time".
+            time_attribute (Optional[MedRecordAttribute], optional):  The time
+                attribute. If None, the treatment effect analysis is performed in an
+                static way (without considering time). Defaults to None.
             washout_period_days (Dict[str, int], optional): The washout period in days
-                for each treatment group. Defaults to dict().
+                for each treatment group. In the case of no time attribute, it is not
+                applied. Defaults to dict().
```
LauraBoenchenLB, line 122: Maybe we need a warning in the builder if a washout or grace period is set but no time attriubute? Otherwise the users may expect their settings to be applied even if they are not.

LauraBoenchenLB (changes requested):
Great implementation! Needs a few tweaks here and there, but the basis will be good for simple TEEs

medmodels/treatment_effect/continuous_estimators.py
MarIniOnz: It is retrieving all edges that connect those nodes in the control outcome true set with the outcome. It is not checking whether they are connected, but retrieving the edges that connect them!

medmodels/treatment_effect/treatment_effect.py
MarIniOnz: Good point!

medmodels/treatment_effect/treatment_effect.py
```diff
@@ -198,7 +204,17 @@ def _find_groups(
...
         treated_set = self._find_treated_patients(medrecord)
-        treated_set, washout_nodes = self._apply_washout_period(medrecord, treated_set)
+
+        if self._time_attribute:
+            treated_set, washout_nodes = self._apply_washout_period(
+                medrecord, treated_set
+            )
+        else:
+            logging.warning(
+                "Washout period is not applied because the time attribute is not set."
```
LauraBoenchenLB, line 214: This warning should only appear if a washout period is set

medmodels/treatment_effect/treatment_effect.py
LauraBoenchenLB: Maybe it's possible to give this warning in the `_set_configuration() ` Function?

medmodels/treatment_effect/continuous_estimators.py
```diff
@@ -15,7 +16,7 @@ def average_treatment_effect(
     outcome_group: Group,
     outcome_variable: MedRecordAttribute,
     reference: Literal["first", "last"] = "last",
-    time_attribute: MedRecordAttribute = "time",
+    time_attribute: Optional[MedRecordAttribute] = "time",
```
LauraBoenchenLB, line 19: Here defaults to non-static?

LauraBoenchenLB (changes requested):
Just one more request if possible :)

LauraBoenchenLB (approved):
Great job! I think all cases are handled well now in the tests and the static option will make a simple analysis much easier! 🦭

medmodels/treatment_effect/treatment_effect.py
```diff
@@ -113,12 +117,12 @@ def _set_configuration(
...
         outcome: Group,
         patients_group: Group = "patients",
-        time_attribute: MedRecordAttribute = "time",
+        time_attribute: Optional[MedRecordAttribute] = None,
         washout_period_days: Optional[Dict[Group, int]] = None,
         washout_period_reference: Literal["first", "last"] = "first",
         grace_period_days: int = 0,
         grace_period_reference: Literal["first", "last"] = "last",
-        follow_up_period_days: int = 365,
+        follow_up_period_days: int = 1000 * 365,
```
JabobKrauskopf, line 125: confirm that the default folow up period should be 1000 years?

MarIniOnz: Yes, I wanted to set it to `inf` but I needed to set an `int`

JabobKrauskopf: you could use `sys.maxsize` to symbolize a very very large `int`. But `1000` years would also be fine imo

MarIniOnz:
```
>       return self._medrecord.select_nodes(
            lambda node: query(NodeOperand._from_py_node_operand(node))
        )
E       pyo3_runtime.PanicException: Call must succeed: PyErr { type: <class 'OverflowError'>, value: OverflowError('Python int too large to convert to C int'), traceback: Some(<traceback object at 0x1655d8480>) }

medmodels/medrecord/medrecord.py:1273: PanicException
```
`sys.maxsize` did not work with queries

LauraBoenchenLB (approved):

medmodels/treatment_effect/builder.py
```diff
@@ -93,12 +93,13 @@ def with_patients_group(self, group: Group) -> TreatmentEffectBuilder:
         return self
 
     def with_time_attribute(
-        self, attribute: MedRecordAttribute
+        self, attribute: Optional[MedRecordAttribute]
```
JabobKrauskopf, line 96: imo this should not be optional. If a user wants to create a treatment effect using the builder without a time attribute, he should just not call this function

JabobKrauskopf: It might be good to include this in the general docstrings of the Treatment effect builder though. Something along the lines of "by default, it uses static analysis, if you want temporal analysis, you have to define the time attribute using `with_time_attribute`".

JabobKrauskopf (changes requested):
Only super minor stuff, will approve when fixed

JabobKrauskopf (approved):
Screen Shot 2025-01-15 at 12 41 22

PR #297, opened 2025-01-14
refactor: pin polars to 0.45.0 (rust), 1.17.0 (python)

JabobKrauskopf:
This PR pins the polars to:
- `0.45.0` rust version
- `1.17.0` python version

Technically pinning the versions is not necessary as the python polars to rust polars conversion via `pyo3-polars` is actually compatible with most versions at the loss of efficiency. Nonetheless, I think this should be good practice.

Some code changes were required for the update, including rechunking dataframes (when necessary).

---

**Developer Notice**
You might need to update rust in order to comply with the updated `Cargo.lock`. To do this, run `rustup update`


MarIniOnz (approved):
DALLE2025-01-1511 22 37-Acheerfulsealgivingathumbs-upgesturewithahairstyleresembling1980scurlyblackhair Thesealhasahappyexpressionandisdepictedinavi-ezgif com-webp-to-jpg-converter

PR #298, opened 2025-01-14
fix: add polars support for all datatypes

JabobKrauskopf:
Can be merged after #297 

The "culprit" for the bug can be found in: https://github.com/pola-rs/polars/blob/main/crates/polars-core/src/series/from.rs#L208
The conversion from arrow arrays to polars series is only enabled for some datatypes by default. The others need to be enabled using feature flags.

---

**Note:**
We will not include any tests for this behaviour as this would test behaviour that was not written by us, but by the `polars` team. To achieve theoretical test coverage on our side, we'd need to test every permutation of every data-type and check if the dataframe can be transferred from python to rust "space", which would use their code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -428,6 +428,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f8eb564c5c7423d25c886fb561d1e4ee69f72354d16918afa32c08811f6b6a55"
 
 [[package]]
+name = "float-cmp"
+version = "0.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b09cf3155332e944990140d967ff5eceb70df778b34f77d8075db46e4704e6d8"
+dependencies = [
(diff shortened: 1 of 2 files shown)


MarIniOnz:
> The "culprit" for the bug can be found in: https://github.com/pola-rs/polars/blob/main/crates/polars-core/src/series/from.rs#L208

damn you mischievous culprit

MarIniOnz (changes requested):
Sorry to be this guy, for looks to me like we are not proving here the support is added, might need some tests to prove so, or add a ticket that requires that

MarIniOnz (approved):
After answering in a Note in the description of PR to my previous comment, APPROVED.

DALLE2025-01-1512 18 06-AwhimsicalillustrationofasealhuggingapolarbearinasnowyArcticsetting Thesealandthepolarbeararedepictedwithjoyfulexpressionsc-ezgif com-webp-to-jpg-converter

PR #299, opened 2025-01-15
chore: Drop support for python 3.9 and add support for 3.13

JabobKrauskopf:
Removed python 3.9 from the supported list and refactored code to no longer inclue the sys.version checks. Also added support for python 3.13

### DIFF
--- a/.github/workflows/formatting.yml
+++ b/.github/workflows/formatting.yml
@@ -9,10 +9,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - name: Set up Python 3.9
+      - name: Set up Python 3.10
         uses: actions/setup-python@v5
         with:
-          python-version: 3.9
+          python-version: "3.10"
       - name: Set up Rust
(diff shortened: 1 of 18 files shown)


.github/workflows/formatting.yml
```diff
@@ -9,10 +9,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - name: Set up Python 3.9
+      - name: Set up Python 3.10
         uses: actions/setup-python@v5
         with:
-          python-version: 3.9
+          python-version: "3.10"
```
MarIniOnz, line 15: Why now inside quoting marks?

MarIniOnz (commented):
very well done. one small question.

MarIniOnz (approved):
DALLE2025-01-1517 29 38-Adramaticandsurrealsceneofasealstandingonacliffwithanevilsmileusingitsflippertothrowafrightenedsnakeoffthecliff Thesnake-ezgif com-webp-to-jpg-converter

PR #313, opened 2025-01-27
fix: attribute extraction without polars

LauraBoenchenLB:
Extracts attributes from medrecord in lists, so it can handle datatype attribute mismatches directly. 
Before there was a bug because of schema restrictions with polars.

### DIFF
--- a/medmodels/medrecord/_overview.py
+++ b/medmodels/medrecord/_overview.py
@@ -6,8 +6,6 @@ import copy
 from datetime import datetime
 from typing import TYPE_CHECKING, Dict, List, Literal, Optional, Union
 
-import polars as pl
-
 from medmodels.medrecord.schema import AttributesSchema, AttributeType
 from medmodels.medrecord.types import (
     Attributes,
@@ -52,14 +50,17 @@ def extract_attribute_summary(
         Dict[MedRecordAttribute, Union[TemporalAttributeInfo, NumericAttributeInfo,
             StringAttributeInfo]: Summary of node or edge attributes.
     """
-    data = pl.DataFrame(data=[{"id": k, **v} for k, v in attribute_dictionary.items()])
+    list_of_dicts = [{**v} for v in attribute_dictionary.values()]
+
+    data = {}
+    for dictionary in list_of_dicts:
+        for key, value in dictionary.items():
+            data.setdefault(key, []).append(value)
 
     data_dict = {}
 
-    attributes = sorted([col for col in data.columns if col != "id"])
-
-    for attribute in attributes:
-        attribute_values = data[attribute].drop_nulls()
+    for attribute in sorted(data.keys()):
+        attribute_values = [value for value in data[attribute] if value is not None]
 
         if len(attribute_values) == 0:
             attribute_info = {"type": "-", "values": "-"}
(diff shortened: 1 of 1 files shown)


medmodels/medrecord/_overview.py
```diff
@@ -52,14 +50,17 @@ def extract_attribute_summary(
         Dict[MedRecordAttribute, Union[TemporalAttributeInfo, NumericAttributeInfo,
             StringAttributeInfo]: Summary of node or edge attributes.
     """
-    data = pl.DataFrame(data=[{"id": k, **v} for k, v in attribute_dictionary.items()])
+    list_of_dicts = [{**v} for v in attribute_dictionary.values()]
 
-    data_dict = {}
+    data = {}
+    for dictionary in list_of_dicts:
```
MarIniOnz, line 56: There should no need of assigning that variable, right? 
```suggestion
    data = {}
    for dictionary in attribute_dictionary.values():
```

medmodels/medrecord/_overview.py
```diff
@@ -52,14 +50,17 @@ def extract_attribute_summary(
...
+        for key, value in dictionary.items():
+            data.setdefault(key, []).append(value)
 
-    attributes = sorted([col for col in data.columns if col != "id"])
+    data_dict = {}
 
-    for attribute in attributes:
-        attribute_values = data[attribute].drop_nulls()
+    for attribute in sorted(data.keys()):
```
MarIniOnz, line 62: It should work the same if using:
```suggestion
    for attribute in sorted(data):
```

medmodels/medrecord/_overview.py
```diff
@@ -52,14 +50,17 @@ def extract_attribute_summary(
...
+        attribute_values = [value for value in data[attribute] if value is not None]
 
         if len(attribute_values) == 0:
```
MarIniOnz, line 65: Maybe it looks better this way:
```suggestion
        if not attribute_values:
```
@JabobKrauskopf ?

medmodels/medrecord/_overview.py
```diff
@@ -52,14 +50,17 @@ def extract_attribute_summary(
...
             attribute_info = {"type": "-", "values": "-"}
```
MarIniOnz, line 66:
```suggestion
            attribute_info = {"type": "-", "values": "-"}
            
```

medmodels/medrecord/_overview.py
```diff
@@ -71,19 +72,19 @@ def extract_attribute_summary(
                 attribute_info = _extract_temporal_attribute_info(attribute_values)
             else:
                 attribute_info = _extract_string_attribute_info(
-                    attribute_series=attribute_values,
+                    attribute_values=attribute_values,
                     long_string_suffix="categories",
                     short_string_prefix="Categories",
                 )
         # Without Schema
```
MarIniOnz, line 79:
```suggestion

        # Without Schema
```

MarIniOnz (changes requested):
Very minor changeSSSS before a Seal is drawn out of the sea for you!

medmodels/medrecord/_overview.py
LauraBoenchenLB: True!

MarIniOnz (approved):
Gemini_Generated_Image_mwymagmwymagmwym

medmodels/medrecord/_overview.py
JabobKrauskopf: imo `len(attribute_values) == 0` is more expressive

medmodels/medrecord/_overview.py
```diff
@@ -52,17 +50,20 @@ def extract_attribute_summary(
...
-    attributes = sorted([col for col in data.columns if col != "id"])
+    data_dict = {}
 
-    for attribute in attributes:
-        attribute_values = data[attribute].drop_nulls()
+    for attribute in sorted(data):
+        attribute_values = [value for value in data[attribute] if value is not None]
 
-        if len(attribute_values) == 0:
+        if not attribute_values:
```
JabobKrauskopf, line 64: see https://github.com/limebit/medmodels/pull/313#discussion_r1939118558

JabobKrauskopf (commented):
One smol comment. Can also approve, if requested

JabobKrauskopf (approved):
image

MarIniOnz (approved):

PR #317, opened 2025-01-27
fix: example dataset data types

JabobKrauskopf:
The datatypes for the example datasets were not set correctly by polars. This could have been caught by the MedRecord schema. However the schema is set without checking nodes/edges for performance reasons (could have used the `update_schema` function but this would be very inefficient as all nodes/edges would be tested for every medrecord that is created using `MedRecord.from_example_dataset()`).
I have added a test that checks if the schema can be updated without errors.

### DIFF
--- a/crates/medmodels-core/src/medrecord/example_dataset/mod.rs
+++ b/crates/medmodels-core/src/medrecord/example_dataset/mod.rs
@@ -1,7 +1,251 @@
 use super::{datatypes::DataType, AttributeType, GroupSchema, MedRecordAttribute, Schema};
 use crate::MedRecord;
-use polars::{io::SerReader, prelude::CsvReadOptions};
-use std::{collections::HashMap, io::Cursor};
+use polars::{
+    io::SerReader,
+    prelude::{CsvReadOptions, DataType as PolasrDataType, Schema as PolarsSchema, TimeUnit},
+};
+use std::{collections::HashMap, io::Cursor, sync::Arc};
+
+macro_rules! simple_dataset_schema {
+    () => {
+        Schema {
+            groups: HashMap::from([
+                (
+                    "diagnosis".into(),
+                    GroupSchema {
+                        nodes: HashMap::from([("description".into(), DataType::String.into())]),
+                        edges: HashMap::new(),
+                        strict: Some(true),
+                    },
+                ),
+                (
+                    "drug".into(),
+                    GroupSchema {
+                        nodes: HashMap::from([("description".into(), DataType::String.into())]),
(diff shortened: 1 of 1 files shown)


crates/medmodels-core/src/medrecord/example_dataset/mod.rs
```diff
@@ -96,16 +357,35 @@ impl MedRecord {
             .collect::<Vec<_>>();
 
         let cursor = Cursor::new(SIMPLE_PATIENT_DIAGNOSIS);
+        let mut patient_diagnosis_schema = PolarsSchema::with_capacity(4);
+        patient_diagnosis_schema.insert("patient_id".into(), PolasrDataType::String);
+        patient_diagnosis_schema.insert("diagnosis_code".into(), PolasrDataType::String);
+        patient_diagnosis_schema.insert(
+            "time".into(),
+            PolasrDataType::Datetime(TimeUnit::Microseconds, None),
```
LauraBoenchenLB, line 365: Diagnosis time is just a date - does this matter for the polars data type?

LauraBoenchenLB (commented):
Great job and smart idea to test for this schema specifically without compromising on speed for usual medrecord. We have to keep in mind that the schema will not be checked when updated, maybe this can be added at a later point when it is feasible speedwise. 
One small question regarding data type.

crates/medmodels-core/src/medrecord/example_dataset/mod.rs
JabobKrauskopf: It would be cleaner to use Date, however we currently don't support neither polars Date nor `datetime.date`. This will be switched over once we do.

JabobKrauskopf:
@LauraBoenchenLB

I'm not entirely sure what you meant with `We have to keep in mind that the schema will not be checked when updated, maybe this can be added at a later point when it is feasible speedwise.` so this explanation might not be applicable:

When the schema of any (including this) medrecord is updated, it checks if the data in the medrecord fits the new schema.

This is the ONLY exception. Because we have access to the private internals of the medrecord, we are able to set the schema wihtout checking if the data fits. This is done so the schema is not checked every time an example medrecord is created using `from_{simple, advanced}_medrecord` as the call would be 3x slower otherwise (I don't actually know how much slower). But because we (now) have a test that checks if the schema could be properly updated, we can be sure that the schema we configured is correct. The only way this could go wrong is for polars to fuck up spectacularly :D

LauraBoenchenLB (approved):
Perfection ✨ After your previous explanation even better, I just misunderstood the initial description. Nothing to keep in mind then! 🦭

PR #318, opened 2025-01-28
docs: add version switcher

FloLimebit:
This branch is adding the version switcher functionality to our docs. Additionally, a red warning banner is shown, when the user is viewing old docs versions. 

When rendering the docs locally, the switch works. However, when selecting a version from the switch, the docs are referring you to the actual deployed version of the current docs - at this point the version switch doesn't work anymore. 

Deployment info: 

* To make this version switcher work with previously deployed versions of the docs, the previous docs need to get rerendered
* As discussed with @JabobKrauskopf, rerendering the docs should be a one time manual task since the changes just affect some minor front-end functionalities and don't change any content

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -156,7 +156,7 @@ html_theme_options = {
         "json_url": "https://www.medmodels.de/docs/switcher.json",
         "version_match": version,
     },
-    "check_switcher": False,
+    "show_version_warning_banner": True,
 }
 
 html_context = {
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,20 +1,20 @@
 """Sphinx configuration."""
 
+import importlib.metadata
 import sys
 from datetime import date
 from pathlib import Path
 
-from myst_parser import __version__
 from sphinx.application import Sphinx
 
 # Add parent directory to sys.path for autodoc
 sys.path.insert(0, str(Path(__file__).parent.parent))
 
 # Project information
-project = "medmodels"
+project = "MedModels"
 author = "Limebit GmbH"
 copyright = f"{date.today().year}, {author}"
-version = __version__
+version = importlib.metadata.version(project)
 
 # General configuration
 extensions = [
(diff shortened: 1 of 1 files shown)


JabobKrauskopf (approved):
Screen Shot 2025-01-28 at 11 53 08

FloLimebit:
> It does not work properly yet.
> 
> You get this when you open the dosc-serve direction
> 
> * choose version works, but it shows this weird switch to stable version banner. Is that intended?
>   image
> * Also, when a version is chosen, the choose version button does not work anymore. Also intended?
>   image


@MarIniOnz 

This is expected behavior: 

> When rendering the docs locally, the switch works. However, when selecting a version from the switch, the docs are referring you to the actual deployed version of the current docs - at this point the version switch doesn't work anymore.

The warning banner pops up, because the locally rendered version does not match a latest deployed version. Sphinx is checking [this](https://www.medmodels.de/docs/switcher.json) file to validate whether the currently viewed docs version is the latest version. If not, the warning banner pops up.

MarIniOnz:
> > It does not work properly yet.
> > You get this when you open the dosc-serve direction
> > 
> > * choose version works, but it shows this weird switch to stable version banner. Is that intended?
> >   image
> > * Also, when a version is chosen, the choose version button does not work anymore. Also intended?
> >   image
> 
> @MarIniOnz
> 
> This is expected behavior

Is it possible to remove the `Choose version` button once a version is selected? It is a bit misleading. Or is it just not possible?

FloLimebit:
> Is it possible to remove the `Choose version` button once a version is selected? It is a bit misleading. Or is it just not possible?

It will show the chosen version number when properly deployed, not the "Choose version" string - this is only happening because local deployment and server side deployment are not in sync. 

We are using the Pydata Sphinx Theme which also provides the version switch function. You can check the real behaviour of the switch on their docs page for example: https://pydata-sphinx-theme.readthedocs.io/en/stable/   

This is how our version switch will act as well as soon as it is deployed to the live docs

MarIniOnz (approved):
DALLE2025-01-2912 01 45-Afuturisticmachineresemblingacharactercreatorscreenfromavideogame Themachinehasseveralcircularscreensdisplayingdifferentversionso-ezgif com-webp-to-jpg-converter

PR #320, opened 2025-01-29
fix: adding a v to the version number via f-string

FloLimebit:
When sphinx generates the docs it is adding a "v" to the derived version number now


JabobKrauskopf (approved):
image

Issue #321, opened 2025-02-03
Allow for errors to be thrown and shown in documentation


JabobKrauskopf:
Extend the sphinx exension to support throwing errors. A distinction between accidental errors and intended errors needs to be made

PR #324, opened 2025-02-05
docs: allow for errors to be thrown and shown in documentation

MarIniOnz:
# Medmodels Pull Request

## Description

Adding an `Sphinx` extension that allows for showing errors when they are intended. Closes Issue #321 .

## Checklist

- [X] This Pull Request follows the rules established in the [Developer Guide](https://www.medmodels.de/docs/latest/developer_guide/pull-request.html).

### DIFF
--- a/docs/sphinx_extensions.py
+++ b/docs/sphinx_extensions.py
@@ -1,11 +1,21 @@
 """Sphinx extension to execute included code snippets and display their output.
 
-This module defines a custom Sphinx directive `ExecLiteralInclude` that allows
-including code from external files, executing it, and displaying both the code
-and its output in the documentation.
+This module defines two custom Sphinx directives:
+1. `exec-literalinclude` that allows including code from external files,
+   executing it, and displaying both the code and its output.
+2. `exec_error-literalinclude` that executes code and, if an error occurs,
+   prints the full traceback along with the code.
 
 Example:
     ```{exec-literalinclude} path/to/your_script.py
+    ---
+    language: python
+    setup-lines: 1-4
+    lines: 6-10
+    ---
+    ```
+
+    ```{exec_error-literalinclude} path/to/your_script_with_error.py
     ---
     language: python
     setup-lines: 1-4
@@ -24,6 +34,41 @@ from docutils import nodes
 from sphinx.application import Sphinx
 from sphinx.util import parselinenos
 from sphinx.util.docutils import SphinxDirective
+
+
+class ErrorMessageNode(nodes.General, nodes.Element):
+    """A custom node to represent a formatted error message."""
+
+    def __init__(self, message: str) -> None:
+        """Initialize the ErrorMessageNode.
+
+        Args:
+            message (str): The error message to display.
+        """
+        super().__init__()
+        self.message = message
+
+
+def visit_error_message_node(self: nodes.NodeVisitor, node: ErrorMessageNode) -> None:
+    """Visitor function to generate HTML for ErrorMessageNode.
+
+    Args:
+        self (nodes.NodeVisitor): The visitor instance.
+        node (ErrorMessageNode): The ErrorMessageNode instance.
+    """
+    self.body.append(  # pyright: ignore[reportAttributeAccessIssue]
+        f'<div class="admonition error"><p class="admonition-title error-title">Error:</p><p>{node.message}</p></div>'
+    )
+
+
+def depart_error_message_node(self: nodes.NodeVisitor, node: ErrorMessageNode) -> None:
+    """Departure function for ErrorMessageNode.
+
+    Args:
+        self (nodes.NodeVisitor): The visitor instance.
+        node (ErrorMessageNode): The ErrorMessageNode instance.
+    """
+    pass
 
 
 class ExecLiteralInclude(SphinxDirective):
@@ -46,9 +91,9 @@ class ExecLiteralInclude(SphinxDirective):
                 and its output.
 
         Raises:
-            FileNotFoundError: If the specified file does not exist.
-            Exception: If an error occurs while executing the code.
-        """  # noqa: DOC502
+            KeyboardInterrupt: If the code raises a KeyboardInterrupt exception.
+            SystemExit: If the code raises a SystemExit exception.
+        """
         environment = self.state.document.settings.env
         _, filename = environment.relfn2path(self.arguments[0])
 
@@ -59,7 +104,6 @@ class ExecLiteralInclude(SphinxDirective):
             error = self.state_machine.reporter.error(
                 f"File not found: {filename}", line=self.lineno
             )
-
             return [error]
 
         total_lines = len(code_lines)
@@ -71,7 +115,6 @@ class ExecLiteralInclude(SphinxDirective):
             setup_code = "".join([code_lines[i] for i in setup_line_numbers])
 
         # Extract main code
-        main_code = ""
         if "lines" in self.options:
             main_line_numbers = parselinenos(self.options["lines"], total_lines)
             main_code = "".join([code_lines[i] for i in main_line_numbers])
@@ -84,11 +127,9 @@ class ExecLiteralInclude(SphinxDirective):
 
         # Prepare code for execution
         main_code_lines = main_code.rstrip().split("\n")
-
         # Remove trailing empty lines
         while main_code_lines and not main_code_lines[-1].strip():
             main_code_lines.pop()
-
         if main_code_lines:
             last_line = main_code_lines.pop()
             code_before_last_line = "\n".join(main_code_lines)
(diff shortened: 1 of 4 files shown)


docs/user_guide/02b_query_engine.md
```diff
@@ -149,42 +149,34 @@ lines: 41-56
...
-gender = node.attribute("gender")
-gender.lowercase()
-gender.trim()
-gender.equal_to("m")
+```{exec_error-literalinclude} scripts/02b_query_engine.py
+---
+language: python
+setup-lines: 1-6
+lines: 61-68
+---
```
JabobKrauskopf, line 159: it should be configurable which type of error is expected here. In this case it would be a `PanicException`.
To fulfill the issue requirement
> A distinction between accidental errors and intended errors needs to be made

JabobKrauskopf (changes requested):

docs/sphinx_extensions.py
```diff
@@ -160,10 +200,145 @@ def _parse_line_range(self, line_range_str: str, total_lines: int) -> List[int]:
...
+
+                error_msg = (
+                    f"Unexpected error type. Expected '{expected_error}', "
+                    f"but got '{e.__class__.__name__}'."
+                )
+                return [code_node, ErrorMessageNode(error_msg)]
+
+            # No expected error provided, show full traceback
+            error_msg = f"Error executing code: {type(e).__name__} - {e}"
+            return [code_node, ErrorMessageNode(error_msg)]
```
JabobKrauskopf, line 311: These cases should `raise e`. Otherwise the error will still be in the docs and is not caught during the build process

JabobKrauskopf, line 311: The expected error handling should just be included in `ExecLiteralInclude`. It does not make sense to create a separate directive for this imo.

JabobKrauskopf (changes requested):

docs/sphinx_extensions.py
JabobKrauskopf: Also the error message should just be the actual error message not with the `Expected error occurred: ` prefix

docs/sphinx_extensions.py
```diff
@@ -119,10 +153,20 @@ def run(self) -> List[nodes.Node]:  # noqa: C901
...
+                if e.__class__.__name__ == expected_error:
+                    return [code_node, ErrorMessageNode(f"{e}")]
+                raise
+            raise
+
+        if expected_error:
+            return [
+                code_node,
+                ErrorMessageNode(f"Expected error '{expected_error}' did not occur."),
+            ]
```
JabobKrauskopf, line 169: If the expected error did not occur, this should raise an error instead of the error message being baked into the docs

docs/sphinx_extensions.py
```diff
@@ -26,6 +28,41 @@
...
+def visit_error_message_node(self: nodes.NodeVisitor, node: ErrorMessageNode) -> None:
+    """Visitor function to generate HTML for ErrorMessageNode.
+
+    Args:
+        self (nodes.NodeVisitor): The visitor instance.
+        node (ErrorMessageNode): The ErrorMessageNode instance.
+    """
+    self.body.append(  # pyright: ignore[reportAttributeAccessIssue]
+        f'<div class="admonition error"><p class="admonition-title error-title">Error:</p><p>{node.message}</p></div>'
+    )
```
JabobKrauskopf, line 53:
```suggestion
def visit_error_message_node(visitor: nodes.NodeVisitor, node: ErrorMessageNode) -> None:
    """Visitor function to generate HTML for ErrorMessageNode.

    Args:
        visitor (nodes.NodeVisitor): The visitor instance.
        node (ErrorMessageNode): The ErrorMessageNode instance.
    """
    visitor.body.append(  # pyright: ignore[reportAttributeAccessIssue]
        f'<div class="admonition error"><p class="admonition-title error-title">Error:</p><p>{node.message}</p></div>'
    )
```

docs/sphinx_extensions.py
```diff
@@ -119,10 +153,20 @@ def run(self) -> List[nodes.Node]:  # noqa: C901
...
-        except Exception as e:
-            error_msg = f"Error executing code: {e}"
-            error_node = nodes.error("", nodes.paragraph(text=error_msg))
-            return [code_node, error_node]
+        except (KeyboardInterrupt, SystemExit):
+            raise
+        except BaseException as e:
+            if expected_error:
+                if e.__class__.__name__ == expected_error:
+                    return [code_node, ErrorMessageNode(f"{e}")]
```
JabobKrauskopf, line 161:
```suggestion
                    return [code_node, ErrorMessageNode(f"{expected_error}: {e}")]
```

docs/sphinx_extensions.py
```diff
@@ -167,3 +211,7 @@ def setup(app: Sphinx) -> None:
         app (Sphinx): The Sphinx application instance.
     """
     app.add_directive("exec-literalinclude", ExecLiteralInclude)
+    app.add_node(
+        ErrorMessageNode, html=(visit_error_message_node, depart_error_message_node)
```
JabobKrauskopf, line 215: This can be replaced with

```suggestion
        ErrorMessageNode, html=(visit_error_message_node, None)
```

Then the `depart_error_message_node` can be removed

docs/user_guide/02b_query_engine.md
```diff
@@ -149,42 +149,36 @@ lines: 41-56
 :::
 
 :::{note}
-Query methods used for changing the operands cannot be concatenated or assigned to variables, since their Return is None. That is, the following code snippet will set `gender_lowercase` as None, and as a result, an AttributeError will be thrown:
+Query methods used for changing the operands cannot be assigned to variables, for further querying, since their _Return_ is `None`. That is, the following code snippet will set `gender_lowercase` as None, and as a result, an `AttributeError` will be thrown:
```
JabobKrauskopf, line 152: check the commas here. I think there might be too many

JabobKrauskopf, line 152:
```suggestion
Query methods used for changing the operands cannot be assigned to variables, for further querying, since their return type is `None`. That is, the following code snippet will set `gender_lowercase` as None, and as a result, an `AttributeError` will be thrown:
```

JabobKrauskopf: The `that is` sounds a bit weird at the beginning. You didn't change it but you're already here

JabobKrauskopf (changes requested):
Teeny tiny smol stuff

docs/sphinx_extensions.py
MarIniOnz: Tried this and it does not work, it requires a function like that one.

JabobKrauskopf: Okay 👍
Then the `self` in the `depart_error_message_node ` should also be renamed to `visitor`

docs/sphinx_extensions.py
```diff
@@ -1,8 +1,10 @@
 """Sphinx extension to execute included code snippets and display their output.
 
-This module defines a custom Sphinx directive `ExecLiteralInclude` that allows
-including code from external files, executing it, and displaying both the code
-and its output in the documentation.
+This module defines two custom Sphinx directives:
+1. `exec-literalinclude` that allows including code from external files,
+   executing it, and displaying both the code and its output.
+2. `exec_error-literalinclude` that executes code and, if an error occurs,
+   prints the full traceback along with the code.
```
JabobKrauskopf, line 7: Needs to be updated

docs/user_guide/02b_query_engine.md
```diff
@@ -149,42 +149,36 @@ lines: 41-56
 :::
 
 :::{note}
-Query methods used for changing the operands cannot be concatenated or assigned to variables, since their Return is None. That is, the following code snippet will set `gender_lowercase` as None, and as a result, an AttributeError will be thrown:
+Query methods used for changing the operands cannot be assigned to variables for further querying, since their return type is `None`. That can be appreciated in the following code snippet, where the variable `gender_lowercase` in the query takes None as value. An `AttributeError` is thrown as a consequence when trying to further query with the `equal_to` querying method:
```
JabobKrauskopf, line 152:
```suggestion
Query methods used for changing the operands cannot be assigned to variables for further querying, since their return type is `None`. The following code snippet shows an example, where the variable `gender_lowercase` evaluates to None. An `AttributeError` is thrown as a consequence when trying to further query with the `equal_to` querying method:
```

JabobKrauskopf (changes requested):

JabobKrauskopf (approved):
unnamed

LauraBoenchenLB (approved):
Gemini_Generated_Image_dw02eadw02eadw02

great new feature for the docs, good job! :)

PR #345, opened 2025-03-05
fix: error when node with edges to itself is deleted

JabobKrauskopf:


MarIniOnz (approved):
Gemini_Generated_Image_jvn6zjvn6zjvn6zj

PR #350, opened 2025-03-12
docs: user guide chapter installation

MarIniOnz:


JabobKrauskopf (changes requested):
The information is mostly misleading or incorrect. This should only cover the actual user installation, no dev setup or anything.
IMO the structure should be:
- Prerequisites (only Python 3.10 t o 3.13 - Rust is not required for non-dev installation)
- Installation (we currently only support pypi so give a link to our pip repo and recommend the `pip install medmodels` command)

PR #364, opened 2025-03-25
test: complete test coverage of treatmenteffect

MarIniOnz:


medmodels/treatment_effect/estimate.py
```diff
@@ -532,9 +527,9 @@ def hazard_ratio(self, medrecord: MedRecord) -> float:
             number_control_outcome_true + number_control_outcome_false
         )
 
-        if hazard_control == 0:
-            msg = "Control hazard rate is zero, cannot calculate hazard ratio."
-            raise ValueError(msg)
+        if hazard_control == 0:  # pragma: no cover
+            msg = "Should not be reached"
+            raise NotImplementedError(msg)
```
JabobKrauskopf, line 532: "should never be reached" checks are done due to typing. If this actually can never be 0, this check is not necessary. a comment above `return hazard_treat / hazard_control` saying that `hazard_control` cannot be 0 is sufficient

medmodels/treatment_effect/treatment_effect.py
```diff
@@ -584,16 +581,19 @@ def _query_node_within_time_window(
...
 
         max_time_window = time_of_treatment.clone()
-        if end_days < 0:
-            max_time_window.subtract(timedelta(-end_days))
-        else:
-            max_time_window.add(timedelta(end_days))
+
+        if end_days < 0:  # pragma: no cover
+            msg = "Should never be reached"
+            raise NotImplementedError(msg)
```
JabobKrauskopf, line 594: see above. comment is enough

JabobKrauskopf (approved):
image

LauraBoenchenLB (approved):
Gemini_Generated_Image_llmjptllmjptllmj

Issue #355, opened 2025-03-24
Improve coverage of MedRecord


JabobKrauskopf:
The medrecord module needs to reach 100% code coverage

PR #375, opened 2025-04-08
feat: implement query return values

JabobKrauskopf:
Important note: This PR does not fix the bug described in #376 so there might still be some issues based on that.

This PR does not include any tests, they will be added later on in #355 

Example Query:

```python
import time

from medmodels.medrecord import MedRecord, NodeOperand

medrecord = MedRecord.from_advanced_example_dataset()


def query(node: NodeOperand):
    node.attribute("age").greater_than(100)

    return node.attribute("age").mean()


start = time.time()
result = medrecord.query_nodes(query)
end = time.time()
print(medrecord)
print(f"Query took {end - start} seconds")
print(f"Query result: {result}")
```


medmodels/medrecord/querying.py
```diff
@@ -27,8 +27,88 @@
...
+    EdgeIndicesQueryResult,
+    EdgeIndexQueryResult,
+    NodeIndicesQueryResult,
+    NodeIndexQueryResult,
+    MultipleValuesQueryResult,
+    SingleValueQueryResult,
+]
+
+NodeQuery: TypeAlias = Callable[["NodeOperand"], QueryReturnOperand]
+NodeQueryExtension: TypeAlias = Callable[["NodeOperand"], None]
```
MarIniOnz, line 104: What is it that you call here a NodeQueryExtension? Is it a query that has no return and it is used as an intermmediate query? 

Maybe better naming... NodeQueryComponent?

medmodels/medrecord/indexers.py
```diff
@@ -515,14 +602,26 @@ def __setitem__(  # noqa: C901
...
+
+                for attribute in attributes:
                     self._medrecord._medrecord.update_node_attribute(
-                        [node], attribute, value
+                        [query_result],
+                        attribute,
+                        value,
                     )
 
             return None
```
MarIniOnz, line 627: Should we not have a warning or Error of some sort in case there is no query_result? (what happens when else: is triggered?)

MarIniOnz (changes requested):
This is one of the most awaited features I have been waiting for. Game changer!

Small couple comments.

medmodels/medrecord/querying.py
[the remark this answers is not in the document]
JabobKrauskopf: `NodeQueryComponent` me likey

medmodels/medrecord/indexers.py
JabobKrauskopf: I also thought about this a lot.

I don't think this should throw an Error (except for when the typing contract would otherwise be broken, e.g. in `__getitem__`). I also don't think this should log a warning (though I'm not 100% sure in this case). I'm thinking that warnings should be used where something caused by the user is happening that is either implicit or that the user cannot really check/comprehend. In this case however, the user could simply execute the query and see that the query matches no nodes.
Also there is nothing "bad" happening here. Exactly the node that is described in the users query (so None) is deleted/updated.

I'm leaning towards not adding a warning

MarIniOnz: Okay, I am convinced. It just felt weird to have and else: clause hanging and did not know whether that was intended.
It is true that not doing anything to the MedRecord should not result in a warning.

MarIniOnz (approved):
Lets get this show going!
ChatGPT Image Apr 10, 2025, 11_55_35 AM

PR #378, opened 2025-04-10
fix: add divide method to `SingleValueOperand`

JabobKrauskopf:


MarIniOnz (approved):
We found it!

ChatGPT Image Apr 10, 2025, 10_58_13 AM

PR #395, opened 2025-04-28
test: restructure test folders and coverage setup

JabobKrauskopf:
I restructured the folder testing structure. instead of tests being located directly where the modules are located, they are added in a separate root `tests` folder. This `tests` folder has the same structure as the `medmodels` folder.

Because we now have a strict structure, I was able to add a script which enforces that coverage of a module is only calculated based on its' exact test (e.g. only `test_medrecord` contributes to `medrecord` coverage).


MarIniOnz (approved):
Poor thing trying to cover his food :3
Gemini_Generated_Image_mx125amx125amx12

